PR #1, opened 2017-12-15
Use github releases to download a specific initial version

bcardiff:
I made a PR to download the binary from github directly. This way it does not depends on which is the stable crystal package published in debian.


RX14:
Thanks!

PR #2, opened 2017-12-18
Use fpm via docker

bcardiff:
I try using ruby:2.4-alpine but the tar in alpine was complaining about --owner=0 option.
I split the cache & pull docker flags from build args
The temp directory now is fixed to /tmp/crystal, I've no strong opinion here, it just felt simpler (meanwhile there was a miss typed $$tmpdir -> $$tempdir in rpm target that now is "fixed")

### DIFF
--- a/x86_64/Dockerfile-fpm
+++ b/x86_64/Dockerfile-fpm
@@ -1,0 +1,8 @@
+FROM ruby:2.4
+
+RUN apt-get update \
+ && apt-get install -y rpm \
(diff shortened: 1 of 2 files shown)


RX14:
Does the `docker-fpm` target run every make invocation?

bcardiff:
@RX14 if no-cache is used, then yes. I couldn't find a better balance. Feel free to change it.

PR #5, opened 2018-02-10
Add targets for docs, osx, docker. Update dependencies. Allow parametrization of build

bcardiff:
This adds targets for: 

* osx (via omnibus) `cd x86_64; make darwin ...`. This targets run on an osx host directly. 
* docker images (by installing linux 64 bits unsigned .deb) `cd docker; make ...`.
* docs (by compiling docs using a docker image with the compiler) `cd docs; make ...`.

All fixed previous crystal version, package iteration, and output crystal version were removed and now should be parametrized by environment variables.

Params for what to build:

* `CRYSTAL_SHA1` specifies what specific commit to build.

Params to choose which artifact with a crystal compiler use:

* `PREVIOUS_CRYSTAL_RELEASE_LINUX_TARGZ` needs to be a *-linux-x86_64.tar.gz url
* `PREVIOUS_CRYSTAL_RELEASE_DARWIN_TARGZ` needs to be a *-darwin-x86_64.tar.gz url, these specifies which released compiler use in the compilation for x86_64 linux and darwin targets.
* `CRYSTAL_DEB` needs to be a .deb file (usually unsigned) and is used in docker image target.
* `CRYSTAL_DOCKER_IMAGE` needs to be an image name 

Params to choose how to brand the outputs:

* `CRYSTAL_VERSION` and `PACKAGE_ITERATION` determines how to brand de produced binaries. There is no need for `CRYSTAL_VERSION` to be a tag. It can be `nightly`.
* `DOCKER_TAG` specifies how to brand the docker images. There are two images built, one with the specified tag and other with `-build` suffix that has llvm 4.0 and is prepared to build the compiler. The docker images are just built, not pushed.


bcardiff:
A preview of how this will be used for nightly / tagged release can be found at https://github.com/crystal-lang/crystal/blob/ci/nightly/.circleci/config.yml

RX14:
First impressions: osx should be in a different folder, I think it would be best to rename `x86_64` to `linux` and `macos`. Docker should be in the Linux makefile.

RX14:
I also want the user experience to be as simple as checking out this repo and running make. 

For that reason, all these new config variables should have defaults. When we release a new crystal version those defaults should be updated.

x86_64/Dockerfile
```diff
@@ -98,7 +100,6 @@ RUN git clone https://github.com/crystal-lang/crystal \
  && make docs \
  && touch src/compiler/crystal.cr \
  && make crystal stats=true static=true ${release:+release=true} \
-                 CRYSTAL_CONFIG_VERSION=${crystal_version} \
```
RX14: Why?

bcardiff:
@RX14 The responsibility of the lineage will be in crystal-lang/crystal repo. Something we think is good is to not need a release/tag in both repo for each release. I don't see how having a default that will get outdated soon adds value.

I agree we can split the `x86_64` in linux / darwin. When if linux 32 bits is added, should it be in linux or a separate so the folders are platform+arch?

x86_64/Dockerfile
bcardiff: Because 

1) for nightly packages we want the branded version as nightly but leave `crystal --version` with more description to target specific commit

2) align the `crystal --version` with the darwin/omnibus output

This implies that the `crystal --version` output will be determined by the code inside the compiler during the make.

RX14:
Except that a true release of crystal depends both on the version of crystal and shards, and libgc. That's why I was tagging each release. If we don't want to go that path, the only configurable should still be `CRYSTAL_VERSION`. I won't accept users having to set up 8+ env vars to reproduce the output. UX is very important to me, even for a build process like this.

- `CRYSTAL_SHA1` should default to `v$(CRYSTAL_VERSION)`
- `PACKAGE_ITERATION` should default to `1`
etc...

I'm not sure about the foldering, we'll see how it goes. I think there will probably be more commonality between linux 64 and 32bit so perhaps we could make it the same folder.

The `docker` folder should merged with `linux`, and depend on the outputted `.deb` properly in make instead of being a separate makefile with an env var. `docs` folder I dislike too (isn't this already automated on travis).

x86_64/Dockerfile
RX14: I disagree, the package version shouldn't just be "nightly". The `crystal --version` should align exactly with the version of the package. We should embed git commit hash into `crystal --version` in addition to the `CRYSTAL_CONFIG_VERSION` and this means distro maintainers automatically upgrade to the better `--version` output.

bcardiff:
I split the darwin and linux scripts.

I think there are some different perspectives on how this will be used. 

1. I don't think this repo main goal is to allow users build from sources the compiler. I think this is an artifact to automate and simplify the build process. 
2. Packaging script would probably not change on every release.
3. I see no point in requiring to git tag two repositories on every single release of the compiler. It's a source of ambiguity.

In the ci scripts the tasks are run in different stages, passing artifacts from one to another. That is why the docker target receive it's input and is not just another step in the linux scripts.

Keeping which is the previous compiler and which is the distribution script release used for a release will be available right in the tagged commit in the crystal. All the information in a single place. That is why I thing that allowing parameters is the way to go.

I see no down side in adding a script to default values of the variables to the latest release of the compiler and master branch of the compiler for example. But that script should invoke the makefiles with the parameters determined at the moment of execution. Putting a fixed crystal version that will be outdated soon in this repo does not seems right to me.

bcardiff:
The package iteration used by the ci is now always 1, because actually is not overwritten.

The `CRYSTAL_VERSION` will be
* $CIRCLE_TAG for tagged releases
* nightly-$(date '+%Y%m%d') for nightly
* There is also a workflow for branches named `release/VERSION` and `\bci\b`(containing the word ci) that will use VERSION or the whole name of the branch, in both cases with `-dev` suffix.

I still hold that it's cleaner that the docker and docs folder to be as it is since it is used as different jobs in the ci.

PR #9, opened 2018-05-28
Feature/linux 32 musl

bcardiff:
Automate linux 32 bits binaries as in 64 bits. Even use the same Dockerfile to keep packages layout in sync.

* `$ make all` will build 64 and 32 bits packages
* `$ make all64` will build 32 bits packages
* `$ make all32` will build 32 bits packages

This procedure was already checked in circleci at https://circleci.com/gh/crystal-lang/crystal/6798#artifacts/containers/0


linux/Makefile
```diff
@@ -33,12 +33,12 @@ DEB_NAME = crystal_$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION)_amd64.deb
 RPM_NAME = crystal-$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION).x86_64.rpm
 
 DEB_NAME32 = crystal_$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION)_i386.deb
-RPM_NAME32 = crystal_$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION).i386.rpm
+RPM_NAME32 = crystal-$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION).i386.rpm
```
RX14: This was to conform with the standard rpm naming conventions.

linux/Makefile
```diff
@@ -33,12 +33,12 @@ DEB_NAME = crystal_$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION)_amd64.deb
...
 
 DOCKER_BUILD_ARGS = $(if $(no_cache),--no-cache )$(if $(pull_images),--pull )
 BUILD_ARGS_COMMON = $(DOCKER_BUILD_ARGS)$(if $(release),--build-arg release=true )--build-arg crystal_version=$(CRYSTAL_VERSION) --build-arg crystal_sha1=$(CRYSTAL_SHA1) --build-arg shards_version=$(SHARDS_VERSION) --build-arg gc_version=$(GC_VERSION) --build-arg libatomic_ops_version=$(LIBATOMIC_OPS_VERSION) --build-arg libevent_version=$(LIBEVENT_VERSION) --build-arg package_iteration=$(PACKAGE_ITERATION)
-BUILD_ARGS = $(BUILD_ARGS_COMMON) --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX_TARGZ)
-BUILD_ARGS32 = $(BUILD_ARGS_COMMON) --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX32_TARGZ)
+BUILD_ARGS = $(BUILD_ARGS_COMMON)   --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX_TARGZ)   --build-arg debian_image=debian:8       --build-arg alpine_image=alpine:3.7       --build-arg musl_target=x86_64-linux-musl  --build-arg gnu_target=x86_64-unknown-linux-gnu
```
RX14: Can't we align by aligning the `=` instead?

linux/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM debian:8 AS debian
+ARG alpine_image
```
RX14: Duplicated `ARG`

bcardiff: I will try to remove the last one, but the one before the first ARG needs to be there. Docker only allows FROM ARG if the arg is before the from, even in multi stage builds, so that is the reason why the alpine_image arg is defined before the first FROM...

bcardiff:
Nitpicks addressed

RX14:
@bcardiff what about the makefile rename? 

> I'd rename all the 64bits makefile targets to end in 64. Then have an all32, all64 and keep all as all targets (32 and 64 bit)

bcardiff:
64 suffix added in all the places except in the environment variable used as parameter.
The updated test is been tested at https://circleci.com/workflow-run/90e76c67-8ebb-4b3e-a616-0c638cf6c3e8

PR #12, opened 2018-06-22
Update to shards v0.8.1

bcardiff:
CircleCI is ready to be happy about it https://circleci.com/workflow-run/e9144688-50a8-4781-8421-20020e83227d

### DIFF
--- a/omnibus/config/software/shards.rb
+++ b/omnibus/config/software/shards.rb
@@ -1,4 +1,4 @@
-SHARDS_VERSION = "0.8.0"
+SHARDS_VERSION = "0.8.1"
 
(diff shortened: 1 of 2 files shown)


omnibus/config/software/shards.rb
```diff
@@ -14,6 +14,10 @@
   source md5: "f0a52e64537ea6267a2006195e818c4d"
 end
 
+version "0.8.1" do
+  source md5: "f5b5108d798b1d86d2b9b45c3a2b5293"
```
RX14: ouch, we're still using `md5`? :P

PR #19, opened 2018-09-27
fix issue https://github.com/crystal-lang/crystal/issues/4680

feniix:
This PR fixes https://github.com/crystal-lang/crystal/issues/4680 and also adds the libz-dev dependency that is used while running `shards build`

### DIFF
--- a/linux/Makefile
+++ b/linux/Makefile
@@ -103,8 +103,8 @@ $(OUTPUT_DIR)/$(DEB_NAME64): docker-fpm $(OUTPUT_BASENAME64).tar
     && fpm --input-type dir --output-type deb \
            --name crystal --version $(CRYSTAL_VERSION) --iteration $(PACKAGE_ITERATION) \
            --architecture x86_64 $(PACKAGE_BRANDING_ARGS) \
-           --depends gcc --depends pkg-config --depends libpcre3-dev --depends libevent-dev --depends libz-dev \
-           --deb-recommends git --deb-recommends libssl-dev \
+           --depends gcc --depends pkgconf --depends libpcre3-dev --depends libevent-dev \
+           --deb-recommends git --deb-recommends libssl-dev --deb-recommends libz-dev \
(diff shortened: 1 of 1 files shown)


bcardiff:
In #16 I thought pkg-config was only needed in 32 bits packages.
Since this PR is adding pkg-config to deb 64 bits, would you mind adding `--depends pkgconfig` (no dash) to rpm 64 bits also?

Why is libz-dev required while running `shards build`? We will also need this on both rpm packages probably.

ysbaddaden:
I don't understand why libz-dev is required, too.

feniix:
Hi folks, it surprised me too, I got this error:

```
Error target confbot failed to compile:
/usr/bin/ld: cannot find -lz (this usually means you need to install the development package for libz)
collect2: error: ld returned 1 exit status
Error: execution of command failed with code: 1: `cc "${@}" -o '/app/bin/confbot'  -rdynamic  -lz `command -v pkg-config > /dev/null && pkg-config --libs --silence-errors libssl || printf %s '-lssl -lcrypto'` `command -v pkg-config > /dev/null && pkg-config --libs --silence-errors libcrypto || printf %s '-lcrypto'` -lpcre -lm -lgc -lpthread /usr/share/crystal/src/ext/libcrystal.a -levent -lrt -ldl -L/usr/lib -L/usr/local/lib`

The command '/bin/sh -c shards build' returned a non-zero code: 1```

I am using https://github.com/mamantoha/crest but it did not look like `libz` is something that is pulled by `crest` (I am a noob with crystal so I may be wrong)

RX14:
@feniix it's an optional dependency of the standard library, required for HTTP.

feniix:
done @RX14

bcardiff:
@feniix I was re-reviewing this since the build of the installer failed. I missed that for .deb packages this PR uses `pkgconf` instead of `pkg-config`. I missed that before. I've already created a branch to use `pkg-config`.

Is there any reason to prefer `pkgconf` over `pkg-config`?

PR #21, opened 2018-10-03
Patch libgc to fix memory allocation on GCF

RX14:


RX14:
No. Patching dependencies is fine, there's no need to delay a fix to "avoid" patching dependencies as if it's a bad thing.

RX14:
well after we update the GC it won't apply...

PR #22, opened 2018-10-10
Use pkg-config instead of pkgconf package

bcardiff:
Ref: https://github.com/crystal-lang/distribution-scripts/pull/19#issuecomment-428331160

Build is already happy at https://circleci.com/workflow-run/435fcdcc-16a0-497b-be2a-c09356eaa61a


RX14:
What was the error using `pkgconf`?

bcardiff:
The .deb is installed directly in the scripts, and the dependency was not satisfied https://circleci.com/gh/crystal-lang/crystal/12757
That could be fixed of course.

But then I get into knowing (the readme of) pkgconf and notice it require some tweaks in the environment to be a replacement for pkg-config. Either env var or symlink needs to be created. Ref: https://github.com/pkgconf/pkgconf.

That bring me to https://github.com/crystal-lang/distribution-scripts/pull/19#issuecomment-428331160 and why one package would be preferred over the other.

RX14:
@bcardiff i guess `pkgconf` isn't available on all supported versions of debian/ubuntu yet.

PR #24, opened 2018-11-26
Update to alpine 3.8

RX14:


linux/Dockerfile
```diff
@@ -55,8 +55,8 @@ FROM ${alpine_image}
 RUN apk add --no-cache \
       # Statically-compiled llvm
       llvm4-dev llvm4-static \
```
j8r: You can use llvm5

RX14: That's a change for another PR.

linux/Dockerfile
```diff
@@ -76,16 +76,6 @@ RUN git clone https://github.com/ivmai/bdwgc \
...
-ARG libevent_version
-RUN git clone https://github.com/libevent/libevent \
- && cd libevent \
- && git checkout ${libevent_version} \
- \
- && ./autogen.sh \
- && ./configure --disable-debug --disable-shared \
- && make -j$(nproc)
-
 ENV LIBRARY_PATH=/bdwgc/.libs/:/libevent/.libs/
```
Sija: Shouldn't `/libevent/.libs/` be dropped from `LIBRARY_PATH`?

RX14: send a PR pls

Sija: done, see #25

PR #28, opened 2019-02-25
[DO NOT MERGE] Remove static linking hack

j8r:
The discussion https://github.com/crystal-lang/crystal/issues/7480 remembered me that this hack is now in mainstream.


straight-shoota:
Eventually we'll want to remove this. But we need a proper solution first.

PR #29, opened 2019-03-08
Add alpine.Dockerfile and refactor docker Makefile

straight-shoota:
With https://github.com/crystal-lang/crystal/pull/7479 merged, the generic linux binaries with gnu default target also work on Alpine linux with musl. So we can use the tar.gz created in the linux build to provide an alpine-based docker image.

Benefits:

* Users can easily build statically linked applications with the official docker image.
* We can use it to build the compiler itself and reduce complexity of the [linux build process](https://github.com/crystal-lang/distribution-scripts/blob/master/linux/Dockerfile).
* We can to run compiler and stdlib specs in circelci on alpine linux.
* The resulting docker image is about half the size of the ubuntu-based image.


docker/Makefile
```diff
@@ -1,29 +1,86 @@
...
+		awk 'BEGIN {FS = ":.*?## "}; {printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2}'
+	@echo
+	@printf '\033[34mconfiguration variables:\033[0m\n'
+	@grep -hE '^[a-zA-Z0-9_-]+ \?=.*?## .*$$' $(MAKEFILE_LIST) |\
+		sort |\
+		awk 'BEGIN {FS = " \\?=.*?## "}; {printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2}'
+	@echo
+	@printf '\033[34mrecipes:\033[0m\n'
+	@grep -hE '^##.*$$' $(MAKEFILE_LIST) |\
+awk 'BEGIN {FS = "## "}; /^## [a-zA-Z_-]/ {printf "  \033[36m%s\033[0m\n", $$2}; /^##  / {printf "  %s\n", $$2}'
```
j8r: Usually when using `awk` can replace `grep`:

```make
.PHONY: help
help: ## Show this help
	@echo
	@printf '\033[34mtargets:\033[0m\n'
	@awk 'BEGIN {FS = ":.*?## "} \
		/^[a-zA-Z0-9_-]+:.*?## .*$$/ { \
		printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 \
		}' $(MAKEFILE_LIST) | sort
	@echo
	@printf '\033[34moptional variables:\033[0m\n'
	@awk 'BEGIN {FS = " \\?=.*?## "} \
		/^[a-zA-Z0-9_-]+ \?=.*?## .*$$/ { \
		printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 \
		}' $(MAKEFILE_LIST) | sort
	@echo
	@printf '\033[34mrecipes:\033[0m\n'
	@awk 'BEGIN {FS = "## "} \
		/^## [a-zA-Z_-]/ {printf "  \033[36m%s\033[0m\n", $$2}; \
		/^##  / {printf "  %s\n", $$2} \
		' $(MAKEFILE_LIST)
```

docker/alpine.Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM alpine:latest as runtime
+
+RUN \
+  apk add --update --no-cache --force-overwrite \
+    # core dependencies
+    gc-dev gcc gmp-dev libatomic_ops libevent-dev musl-dev pcre-dev \
```
j8r: Do we use the `gc-dev` package, that isn't built with `--enable-large-config`?

docker/Makefile
j8r: You have certainly taken this from other `Makefile` in this repo.
Nevermind I can submit a PR for this.

docker/alpine.Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+  ln -s /usr/lib/crystal/bin/crystal /usr/bin/crystal && \
+  ln -s /usr/lib/crystal/bin/shards /usr/bin/shards
+
+CMD ["/bin/sh"]
+
+FROM runtime as build
+
+RUN \
+  apk add --update --no-cache --force-overwrite \
+    llvm-dev llvm-static
```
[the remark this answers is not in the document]
straight-shoota: It could be, but I'd prefer to keep the same format as above.

RX14:
shouldn't you need to rebuild `libcrystal.a`?

docker/alpine.Dockerfile
straight-shoota: All other docker files I've seen use this. But `libgc.a` built with `--enable-large-config` is actually included in the crystal tar.gz.

straight-shoota:
@RX14 No, why? It's included in the tar.gz and should be portable on linux systems. It doesn't matter which libc it will eventually be linked against.

ysbaddaden:
You should recompile it nonetheless, I doubt it's really portable —otherwise we wouldn't have distinct bindings for linux-gnu and linux-musl.

straight-shoota:
If it's not portable, should we distribute it with the generic linux package? And what about the other included binary libraries, `libgc.a` and `llvm_ext.o`?

straight-shoota:
I just verified: `libcrystal.a` is exactly the same, whether it's build on alpine or ubuntu. Same goes for `llvm_ext.o`.

ysbaddaden:
That's the thing: the "generic" linux package *isn't* generic but targets `linux-gnu`.

The provided `libgc.a` is built against glibc, same for `libcrystal.a` (I'm not sure `llvm_ext.o` should be distributed). If a built `libcrystal.a` happens to be identical for both `linux-gnu` and `linux-musl`, then it's just luck, without any guarantee that it will stay that way.

straight-shoota:
Hm okay, so I'll make sure to build them locally, in order to make sure it won't break at some point.

It's not super trivial though, because the Makefile is not included in the binary package for `make clean deps`. Otherwise we need to copy it separately into the docker image, or duplicate the instructions.

`libgc.a` also requires a custom patch. But it's already available for the linux build instructions.

An alternative would be to build the binary libs for musl in the linux build and either include them into the generic linux tar.gz (beside the gnu ones) or provide two separate downloads for musl and gnu.

j8r:
I don't know why a `libgc.a` is shipped – the `libgc.a` that should be used is the one of the OS – problem solved :)

j8r:
For `libcrystal.a`, I agree with @ysbaddaden. However I doubt there is a real need for a dedicated archive for musl.
Usually people on musl are on Alpine or Gentoo, and most through Docker. This OSes are usually up-to-date, they use the package manager to get a ~more/less up-to-date Crystal release.

What I hear the most is people waiting an Alpine Docker image with the latest Crystal release.
Let's do so: we can build Crystal directly in the Dockerfile.

straight-shoota:
We want libgc with `--enable-large-config` and AFAIK `gc-dev` doesn't provide that. That's why we need to build it ourselves.

> Let's do so: we can build Crystal directly in the Dockerfile.

We do. It's just a different dockerfile. I don't think there would be any benefit from duplicating the build.

j8r:
I mean, building Crystal on the same OS as the target – it can be dynamically linked.

straight-shoota:
But why? We need a somewhat recent compiler build anyway in order to that. So we can just use the most recent one and use it without recompiling, because it already works.

bcardiff:
> https://github.com/crystal-lang/distribution-scripts/pull/29#issuecomment-471044881
> I don't know why a libgc.a is shipped – the libgc.a that should be used is the one of the OS – problem solved :)

We do need to ship the `libgc.a`. Shortly we will probably end up linking it statically by default since we will need some custom patch to MT ;-)

---

The files in the `./docker/` folder should only include the package/installer made in `./linux/`. If the `.tar.gz` is able to run in alpine directly then something more like the first commit only is better.

If we can/want to produce a different downloadable asset in the release for alpine that will be available in the github releases, then we should add that process in the `./linux/` and package it in the `./docker`.

bcardiff:
Closed in favor of #44

straight-shoota:
#44 only updates the Dockerfile for building the compiler on alpine. It doesn't provide a Dockerfile for running Crystal on alpine.
These use cases are similar, maybe we can use the same Dockerfile. I'm not sure. There are still different goals and you don't need the entire build environment for a runtime container. And of course we'd need to make the Dockerfile available in the `docker` build workflow.

straight-shoota:
So to move forward with this PR, we could resolve the static libraries issues:

* `libgc.a`: It's best if we use the library provided by the OS. The alpine `crystal` package also depends simply on `gc`. This means we should to get `--enable-large-config` added to `gc` package (for a start, we could do without that, though)
* `libcrystal.a` and `llvm_ext.o` should simply be build using `make deps`. For that we'd just have to import the Makefile into the docker container (either individually or include it in the targz).

Alternatively, we could build upon the linux build Dockerfile from #44. The last steps for packaging the targz would need to be moved to a separate build stage. But then we could use the main part of it as build image, just need to put the crystal files in the appropriate folders in `/usr` to make it usable.
A slimmer runtime image could probably be created by uninstalling the build dependencies specific to the compiler (llvm, maybe some of the build tools).

What do you think?

straight-shoota:
I've updated this branch to install `gc-dev` from apk and build `libcrystal.a` and `llvm_ext.o` using the Makefile which is now included in the tarball.

docker/Makefile
```diff
@@ -1,43 +1,113 @@
...
 OUTPUT_DIR = build
-BUILD_ARGS64 = --build-arg crystal_deb=./tmp/crystal.deb --build-arg base_docker_image=ubuntu:bionic
-BUILD_ARGS32 = --build-arg crystal_deb=./tmp/crystal.deb --build-arg base_docker_image=i386/ubuntu:bionic
+BUILD_CONTEXT_UBUNTU = $(CURDIR)/build-context-ubuntu
+BUILD_CONTEXT_ALPINE = $(CURDIR)/build-context-alpine
+BUILD_ARGS_UBUNTU64 = -f ubuntu.Dockerfile --build-arg crystal_deb=crystal.deb $(BUILD_CONTEXT_UBUNTU) --build-arg base_docker_image=ubuntu:bionic
+BUILD_ARGS_UBUNTU32 = -f ubuntu.Dockerfile --build-arg crystal_deb=crystal.deb $(BUILD_CONTEXT_UBUNTU) --build-arg base_docker_image=i386/ubuntu:bionic
+BUILD_ARGS_ALPINE = -f alpine.Dockerfile --build-arg crystal_targz=crystal.tar.gz $(BUILD_CONTEXT_ALPINE)
+DOCKER_TAG_UBUNTU = $(DOCKER_REPOSITORY):$(DOCKER_TAG)
+DOCKER_TAG_ALPINE = $(DOCKER_REPOSITORY)-alpine:$(DOCKER_TAG)
```
bcardiff: shouldn't it be?

```suggestion
DOCKER_TAG_ALPINE = $(DOCKER_REPOSITORY):$(DOCKER_TAG)-alpine
```

docker/alpine.Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM alpine:3.11 as runtime
+
+RUN \
+  apk add --update --no-cache --force-overwrite \
+    # core dependencies
+    gc-dev gcc gmp-dev libatomic_ops libevent-dev musl-dev pcre-dev \
```
bcardiff: gc-dev is not needed since it is bundled withing the tar.gz.
Or is the embedded gc not compatible with alpine somehow?

Also, note that the embedded has the upstream patch for mt.

linux/Dockerfile
```diff
@@ -123,7 +123,7 @@ RUN \
  \
  # Copy stdlib to /share/crystal/src/
  && mkdir -p /output/share/crystal/ \
- && cp -r /crystal/src /output/share/crystal/src \
+ && cp -r /crystal/src /crystal/Makefile /output/share/crystal/ \
```
bcardiff: I don't follow why you want to ship the Makefile in the distributable tar.gz / .deb packages.

straight-shoota: It's necessary for building libcrystal and llvm_ext. I think it makes sense to distribute it. But there could be other solutions as well. Copying it manually into the docker container besides the tarball is tedious because you need another path option. Having a separate copy which only contains the recipies for these dependenies, either maintained as a copy for the alpine build process or in the main repo would work as well.

bcardiff: But the docker images are for using crystal, not building crystal.
When building crystal the whole git repo will be pulled there.

bcardiff: And the -build images are for having the dependencies to build crystal, once the code is checked out.

docker/Makefile
straight-shoota: Depends what we want. IMO it's cleaner to use different docker repositories for different base images. In a single repo every Crystal release would have a huge number of associated docker tags. Currently it's three operating systems (ubuntu64, ubuntu32, alpine) with two variants each (runtime, build). We don't even provide different OS releases as base images but I think that's already starting to get pretty convoluted. And we'll probably be adding more in the future (windows for example).
I'm fine with changing to a only tag-based single repo nomenclature, but I'd prefer separate repos.

bcardiff: I usually see `-alpine`, like in https://hub.docker.com/_/ruby

docker/alpine.Dockerfile
straight-shoota: The gc bundled with tar.gz should work. But this was actually discussed in this PR. The conclusion was to prefer a build with musl to avoid any incompatibilities. Since `gc-dev` already provides that, it's the easiest solution. The crystal apk package also depends on that.

Yes, we're currently missing out on the upstream patch for MT but as soon as BWDGC gets a new release it's certainly becoming available in apk rather soon. IMO that's good enough for now and saves some hassle.

linux/Dockerfile
straight-shoota: But you need libcrystal for using crystal.

straight-shoota: llvm_ext is not necessary, agreed.

docker/alpine.Dockerfile
straight-shoota: Actually, libgc from the tarball fail now. It used to work before, but I hadn't tested it in a while.
```
/ # crystal eval 'puts "hello"'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: /usr/bin/../lib/crystal/lib/libgc.a(pthread_support.o): in function `GC_thr_init':
pthread_support.c:(.text+0x85e): undefined reference to `gnu_get_libc_version'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: /usr/bin/../lib/crystal/lib/libgc.a(mach_dep.o): in function `GC_with_callee_saves_pushed':
mach_dep.c:(.text+0xb8): undefined reference to `getcontext'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: /usr/bin/../lib/crystal/lib/libgc.a(pthread_start.o): in function `GC_inner_start_routine':
pthread_start.c:(.text+0x51): undefined reference to `__pthread_register_cancel'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: pthread_start.c:(.text+0x78): undefined reference to `__pthread_unregister_cancel'
collect2: error: ld returned 1 exit status
Error: execution of command failed with code: 1: `cc "${@}" -o '/root/.cache/crystal/crystal-run-eval.tmp'  -rdynamic  -lpcre -lm /usr/bin/../lib/crystal/lib/libgc.a -lpthread /usr/share/crystal/src/ext/libcrystal.a -levent -lrt -L/usr/bin/../lib/crystal/lib -L/usr/lib -L/usr/local/lib`
```

bcardiff: If that is the case we need a `linux/Makefile` target that will generate a valid tar.gz for alpine.
That uses as input the previous crystal linux (non-alpine) version. 
Embeds a valid gc for alpine.

Then the -alpine and -alpine-build images would include that .tar.gz.

Once that linux libc -> linux musl build is in place, if wanted, in the future the linux musl can be used to build the next one. But it is not mandatory to start a new lineage.

linux/Dockerfile
straight-shoota: I've changed this to the previous iteration https://github.com/crystal-lang/distribution-scripts/pull/29/commits/30e3ae11240b9fdbd0122260f97823ba0441526e#diff-bd84dfa6214dce9b49812bd18b1e2ca0R42-R47

This way we don't need the Makfile and don't need to have llvm-dev installed.

docker/Makefile
j8r: If `crystal` becomes an Official Docker Image (let's believe :slightly_smiling_face: ), it will become mandatory to use tags to differentiate the different image versions.

docker/alpine.Dockerfile
straight-shoota: Can we move the alpine tarball to a separate PR? I'd really like to move this forward and just have a working alpine docker image. Simply using `gc-dev` from APK seems fine for now.

bcardiff: I was missing the --exclude, I was going to say that the embedded libs should take precedence from the system one.

Ok, we can defer the tar.gz if wanted.

linux/Dockerfile
```diff
@@ -123,7 +123,7 @@ RUN \
  \
  # Copy stdlib to /share/crystal/src/
  && mkdir -p /output/share/crystal/ \
- && cp -r /crystal/src /output/share/crystal/src \
+ && cp -r /crystal/src /output/share/crystal/ \
```
bcardiff:
```suggestion
 && cp -r /crystal/src /output/share/crystal/src \
```

docker/alpine.Dockerfile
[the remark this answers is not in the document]
RX14: Why is `llvm-dev` needed?

docker/Makefile
straight-shoota: Changed to alpine tag.

bcardiff:
Cool, before merging there needs to be a branch named `ci/update` or something that will trigger the maintenance build. 

The `publish_docker` job should be updated to push the -alpine docker image.

Once that is there, we can merge this and update the ci/update with the new master sha1 of distribution-scripts.

docker/alpine.Dockerfile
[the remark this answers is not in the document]
j8r: To compile and link statically crystal programs?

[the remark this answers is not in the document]
RX14: But neither of these are required when the compiler is statically linked.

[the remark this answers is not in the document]
RX14: @straight-shoota did you see this?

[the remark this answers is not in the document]
straight-shoota: Sorry, missed it.
This is the build container. It's used for building the compiler where we need llvm libraries available in order to link them into the compiler executable.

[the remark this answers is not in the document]
RX14: Oh, my bad! I misunderstood. This is fine.

straight-shoota:
Workflow looks good: https://circleci.com/workflow-run/7384ac99-a207-4a98-8c21-038e0710f6c0

Docker images can be tested:
* `crystallang/crystal:ci-test-dev-alpine`
* `crystallang/crystal:ci-test-dev-alpine-build`

bcardiff:
@straight-shoota if you can squash/rebase the `ci/test` branch with the new sha in the main repo we can merge that.

bcardiff:
@straight-shoota I see you updated the ci/test branch but I did not find a PR with it. Did I miss it?

straight-shoota:
Seems like I forgot the PR. Now it's here: crystal-lang/crystal#8708

Issue #33, opened 2019-04-22
Sometimes packages are built without static linking


bcardiff:
0.28.0 32 bits .deb packages went out without static linking.

Yet nightlies worked fine https://circleci.com/gh/crystal-lang/crystal/21793#artifacts/containers/0

The tagged one are not statically linked https://circleci.com/gh/crystal-lang/crystal/21811#artifacts/containers/0

The non-static link also happened in maintenance release ci/preview.

But since last nightlies, before 0.28 worked I thought tagged version would work.

bcardiff:
In the https://circleci.com/gh/crystal-lang/crystal/23243 I included a `ldd` output that shows the compiler is not statically linked.

```
Step 42/43 : RUN ldd /crystal-${crystal_version}-${package_iteration}/lib/crystal/bin/crystal
 ---> Running in cfb289fa3e04
	/lib/ld-musl-x86_64.so.1 (0x7fe912afa000)
	libLLVM-4.0.so => /usr/lib/libLLVM-4.0.so (0x7fe90e879000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7fe90e526000)
	libpcre.so.1 => /usr/lib/libpcre.so.1 (0x7fe90e2cb000)
	libevent-2.1.so.6 => /usr/lib/libevent-2.1.so.6 (0x7fe90e081000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7fe90de6f000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7fe912afa000)
	libffi.so.6 => /usr/lib/libffi.so.6 (0x7fe90dc67000)
	libz.so.1 => /lib/libz.so.1 (0x7fe90da50000)
 ---> 56f0404e09f0
Removing intermediate container cfb289fa3e04
Step 43/43 : RUN ldd /crystal-${crystal_version}-${package_iteration}/lib/crystal/bin/shards
 ---> Running in 34bc5b40004c
	ldd (0x7f9a47082000)
```

bcardiff:
In the https://circleci.com/gh/crystal-lang/crystal/23330 the output is

```
Removing intermediate container 4758144cff46
Step 42/43 : RUN ldd /crystal-${crystal_version}-${package_iteration}/lib/crystal/bin/crystal
 ---> Running in 91564462ee25
	ldd (0x7fb66a066000)
 ---> e76da357088f
[...]

PR #37, opened 2019-05-31
Add ldd output and wait before touch

bcardiff:
Related to #33 .

Sometimes the touch leave the same date as the cc output. The sleep should fix that.
I added ldd output to have some additional manual checks during the build.

I'm not sure if this will solve 100% of the issues, but it reduces the chances and the logs have additional diagnosis information.

Maintenance build at https://circleci.com/workflow-run/e2779fd7-c51c-4b53-9e61-a36db574e779

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -90,12 +90,10 @@ RUN git clone https://github.com/crystal-lang/crystal \
  && git checkout ${crystal_sha1} \
  \
  && make deps \
- && mkdir .build \
- && cc '/crystal-musl.o' -o '.build/crystal' -rdynamic src/llvm/ext/llvm_ext.o `llvm-config --libs --system-libs --ldflags` -lstdc++ -lpcre -lm -lgc -lpthread src/ext/libcrystal.a -levent -lrt \
- && ldd .build/crystal \
- && sleep 2 \
+ && mkdir -p .build/crystal-musl \
+ && cc '/crystal-musl.o' -o '.build/crystal-musl/crystal' -rdynamic src/llvm/ext/llvm_ext.o `llvm-config --libs --system-libs --ldflags` -lstdc++ -lpcre -lm -lgc -lpthread src/ext/libcrystal.a -levent -lrt \
+ && export PATH=.build/crystal-musl/:$PATH \
  \
- && touch src/compiler/crystal.cr \
  && make crystal stats=true static=true ${release:+release=true} \
                  CRYSTAL_CONFIG_TARGET=${gnu_target} \
  && ldd .build/crystal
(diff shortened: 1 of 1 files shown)


RX14:
I'd much prefer to robustly solve this problem by placing the built cross-compiled crystal into PATH, which would solve the problem without hacky delays.

RX14:
Alternatively, see if we can avoid cross-compiling completely now

bcardiff:
I changed it to use a different path. I left the last ldd as a manual check.

Maintenance build at https://circleci.com/workflow-run/d559bd23-1e8a-41e9-8b17-eb15a4b2ec6d

RX14:
Actually, we don't validate the output of `ldd` so I'd prefer it to be removed. Either check the output of `ldd` to make sure it's statically linked and fail the build if it's not, or leave it out and run ldd manually for debugging.

bcardiff:
ok, done.

linux/Dockerfile
```diff
@@ -90,12 +90,13 @@ RUN git clone https://github.com/crystal-lang/crystal \
...
- && cc '/crystal-musl.o' -o '.build/crystal' -rdynamic src/llvm/ext/llvm_ext.o `llvm-config --libs --system-libs --ldflags` -lstdc++ -lpcre -lm -lgc -lpthread src/ext/libcrystal.a -levent -lrt \
+ && mkdir -p .build/crystal-musl \
+ && cc '/crystal-musl.o' -o '.build/crystal-musl/crystal' -rdynamic src/llvm/ext/llvm_ext.o `llvm-config --libs --system-libs --ldflags` -lstdc++ -lpcre -lm -lgc -lpthread src/ext/libcrystal.a -levent -lrt \
+ && export PATH=.build/crystal-musl/:$PATH \
  \
- && touch src/compiler/crystal.cr \
  && make crystal stats=true static=true ${release:+release=true} \
-                 CRYSTAL_CONFIG_TARGET=${gnu_target}
+                 CRYSTAL_CONFIG_TARGET=${gnu_target} \
+ && (if [[ $(ldd .build/crystal | wc -l) -gt "1" ]]; then echo './build/crystal is not statically linked'; ldd .build/crystal; exit 1; fi)
```
j8r:
```suggestion
 && [ "$(ldd .build/crystal | wc -l)" -gt "1" ] && { echo './build/crystal is not statically linked'; ldd .build/crystal; exit 1; }
```
I haven't tested, yet. Better to avoid double brackets, less portable.

bcardiff: The operator was wrong.

I appreciate it, but it does not need to be portable outside alpine docker image and it was already fulfilling the needs.

PR #39, opened 2019-06-13
Add snap target

bcardiff:
The maintenance and tagged builds are already stressed in 

* https://circleci.com/workflow-run/88f4c4a1-7360-4b71-84ff-5cd19e7fb6d0
* https://circleci.com/workflow-run/9a38150d-836b-4a10-9f9e-8c9b0e442478

This will be used to ship nightly builds to snap.

The snap is built by embedding a known tar.gz that can be a url or a local path

### DIFF
--- a/snapcraft/crystal-snap-wrapper
+++ b/snapcraft/crystal-snap-wrapper
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+if [ ! -f $SNAP_USER_COMMON/env-check-success ]; then
+  # check if a simple program can be compiled
+  # if something fails show a banner to the user
+
+  $SNAP/bin/crystal eval 'puts "1"' >/dev/null 2>/dev/null
+  if [ $? -eq 0 ]; then
+    touch $SNAP_USER_COMMON/env-check-success
+  else
+    cat <<EOF
+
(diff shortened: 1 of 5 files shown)


snapcraft/crystal-snap-wrapper
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+if [ ! -f $SNAP_USER_COMMON/env-check-success ]; then
+  # check if a simple program can be compiled
+  # if something fails show a banner to the user
+
```
straight-shoota: Is this the general wrapper which you use to run crystal in snap? Compiling even a little program would delay the actual compiler boot time and make it appear to be slow.
Can't there be a different solution to this? Maybe use the actual error code from the invoked crystal command?

bcardiff: That could be another way to do it but `exec` is used to run the compiler and I don't think something can be done _after_ that.

Using `exec` to delegate is better so the wrapper is more transparent.

As soon as the env is verified once the check will no longer run so is a one time delay.

Unfortunately, there is no support for post-installation messages on snap for now. I would have preferred to do it that way.

PR #44, opened 2019-09-05
Update linux LLVM to 8, and don't build crystal twice

RX14:
I was doing this anyway, and it works on my machine, could you test if this fixes the CI?

### DIFF
--- a/linux/Makefile
+++ b/linux/Makefile
@@ -47,15 +47,15 @@ BUILD_ARGS_COMMON = $(DOCKER_BUILD_ARGS) \
 
 BUILD_ARGS64 = $(BUILD_ARGS_COMMON) \
                --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX64_TARGZ)	\
-               --build-arg debian_image=debian:8 \
-               --build-arg alpine_image=alpine:3.8 \
+               --build-arg debian_image=debian:10 \
+               --build-arg alpine_image=alpine:3.10 \
                --build-arg musl_target=x86_64-linux-musl \
                --build-arg gnu_target=x86_64-unknown-linux-gnu
 
 BUILD_ARGS32 = $(BUILD_ARGS_COMMON) \
                --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX32_TARGZ) \
-               --build-arg debian_image=i386/debian:8 \
-               --build-arg alpine_image=i386/alpine:3.8 \
+               --build-arg debian_image=i386/debian:10 \
+               --build-arg alpine_image=i386/alpine:3.10 \
                --build-arg musl_target=i686-linux-musl \
                --build-arg gnu_target=i686-unknown-linux-gnu
 
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -2,21 +2,10 @@ ARG alpine_image
 ARG debian_image
 FROM ${debian_image} AS debian
 
-RUN apt-get update \
- && apt-get install -y build-essential libevent-core-2.0-5 libevent-dev libevent-extra-2.0-5 libevent-openssl-2.0-5 libevent-pthreads-2.0-5 libpcre3-dev automake libtool pkg-config git software-properties-common apt-transport-https curl \
- && (pkg-config || true)
-
-RUN add-apt-repository "deb http://apt.llvm.org/jessie/ llvm-toolchain-jessie-4.0 main" \
- && curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
+RUN echo "deb http://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list \
  && apt-get update \
- && apt-get install -y llvm-4.0-dev
-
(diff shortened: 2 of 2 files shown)


linux/Makefile
```diff
@@ -47,15 +47,15 @@ BUILD_ARGS_COMMON = $(DOCKER_BUILD_ARGS) \
 
 BUILD_ARGS64 = $(BUILD_ARGS_COMMON) \
                --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX64_TARGZ)	\
-               --build-arg debian_image=debian:8 \
-               --build-arg alpine_image=alpine:3.8 \
+               --build-arg debian_image=debian:10 \
+               --build-arg alpine_image=alpine:edge \
```
bcardiff: We need specific tags here for better reproducibility. `:edge` is not an option.

linux/Makefile
```diff
@@ -47,15 +47,15 @@ BUILD_ARGS_COMMON = $(DOCKER_BUILD_ARGS) \
...
                --build-arg musl_target=x86_64-linux-musl \
                --build-arg gnu_target=x86_64-unknown-linux-gnu
 
 BUILD_ARGS32 = $(BUILD_ARGS_COMMON) \
                --build-arg previous_crystal_release=$(PREVIOUS_CRYSTAL_RELEASE_LINUX32_TARGZ) \
-               --build-arg debian_image=i386/debian:8 \
-               --build-arg alpine_image=i386/alpine:3.8 \
+               --build-arg debian_image=i386/debian:10 \
+               --build-arg alpine_image=i386/alpine:edge \
```
bcardiff: ditto

docker/Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM ubuntu:xenial as runtime
+FROM ubuntu:bionic as runtime
```
j8r: Debian's `buster-slim` image would be more appropriate for Docker, slimmer.

bcardiff: Both are 27mb

j8r: Right, it has changed :+1: 
I remembered slim images was lighter.
So, as you wish then.

linux/Makefile
j8r: Alpine `3.10`

docker/Dockerfile
bcardiff: Yeah. ubuntu images got far better in the last couple of releases. Also, this images need to be aligned with crystal-dist repo were tagged releases are done.

I will actually rollback this change. It needs to be aligned with crystal-dist images for 64 and 32 bits and that means definitely updating to llvm 6 or 8 in linux. I would rather update them after some testing.

@RX14 let's keep xenial here for a little bit longer and upgrade in another PR

linux/Makefile
bcardiff: edge is not 3.10 from docker hub. Was there an explicit need for edge @RX14?

docker/Makefile
```diff
@@ -1,7 +1,7 @@
 CRYSTAL_VERSION ?=   ## How the binaries should be branded
 CRYSTAL_DEB ?=       ## which crystal.deb file to install in the docker images
-DOCKER_TAG ?=        ## How to tag the docker image. a -build will be appended for -build image
-DOCKER_REPOSITORY ?= ## Docker hub repository to commit image
+DOCKER_TAG ?= latest## How to tag the docker image. a -build will be appended for -build image
```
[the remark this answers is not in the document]
RX14: Yes, they should. They're good defaults for local builds. The CI will override them. This change does not change any behaviour that would not already fail.

linux/Makefile
RX14: Yes, `ldd` is broken on 3.10, see https://github.com/crystal-lang/crystal/issues/7946.

We can switch to 3.11 when it's released, depending on `edge` is fine.

docker/Makefile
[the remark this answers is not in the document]
bcardiff: The story of a local build is half way.

To support local builds a more complete solution would be to have a script that will fetch latest release and master's sha1 for example.

But all this is independant of the intention of the PR to update llvm 8 and avoid building crystal twice.

docker/Dockerfile
RX14: If you minsist.

docker/Makefile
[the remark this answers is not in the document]
RX14: It's a harmless change. I'm not going to revert it.

RX14:
Closes #29 btw

linux/Makefile
bcardiff: We can't depend on edge. Using the distribution-scripts in the future will not be as reproducible as using a specific version of the docker image.

Maybe we need to delay this until 3.11 is released and stick with #43

RX14: *sigh*, I'll just patch the ldd script manually

RX14:
okay, there we go

sorry for the force push, accidentallied

bcardiff:
Cool thanks.
Checking the build at https://circleci.com/workflow-run/35cfe1aa-4ce3-466b-a8c4-49a77e4ef610

PR #51, opened 2020-01-11
Enhance Makefiles

Daniel-Worrall:
This PR replaces the duplicate references to the targets and prerequisites with magic variables and reworks the Linux Makefile to handle 32 vs 64 bit dynamically where possible. Some targets were rearranged in the interest of keeping similars together.

I did not attempt handling targets with 32/64 in the variable names as this would have required variable changes and quite a bit of conditionals, though it could be handled.

Feedback highly encouraged.

E: Targets tested to be the same through --dry-run

### DIFF
--- a/linux/Makefile
+++ b/linux/Makefile
@@ -96,6 +96,9 @@ help: ## Show this help
 .PHONY: build
 build: $(OUTPUT_BASENAME64).tar ## Build the raw uncompressed tarball
 
+.PHONY: build32
+build32: $(OUTPUT_BASENAME32).tar ## Build the raw uncompressed tarball
+
 $(OUTPUT_BASENAME64).tar: Dockerfile $(FILES)
 	mkdir -p $(OUTPUT_DIR)
 	docker build $(BUILD_ARGS64) -t crystal-build-temp .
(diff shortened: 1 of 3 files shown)


RX14:
I'm not a fan of the added complexity just to reduce line count. There's no complex logic which is actually deduplicated here. I'm a fan of the `$@` usage change and almost nothing else.

straight-shoota:
Also the change introduces weird doc comments. You do realize that double-hash comments on target lines are used for autogenerating usage documentation?

Using `$@` is great and reordering to group similar recipies is also good.

The generic `%.gz: %` and `%.xz: %` recipies are nice, too. That's actually a simplification.

I think I'm basically fine with everything that doesn't use `SECONDEXPANSION`.

RX14:
If you do reorder, please do it in a seperate commit to the other refactors, otherwise it's difficult to review the diff.

Daniel-Worrall:
Updated, and force pushed.

It should be much easier to review. Took out all the complicated dynamic recipes.

I also tried to format the all64,all32s with dynamic recipes while retaining their comments in the help recipe, but I couldn't come up with a way to process and replace.
```
.PHONY: all64 all32
all64 all32: all%: compress% package% ## Build compressed and distribution packages for % bits
```

I also included a change in the tabular formatting to make whitespace more consistent, but it's rather opinionated.

linux/Makefile
```diff
@@ -100,24 +100,24 @@ $(OUTPUT_BASENAME64).tar: Dockerfile $(FILES)
...
-	  && docker cp "$$container_id":/output/crystal-$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION).tar $(OUTPUT_BASENAME64).tar \
+	  && docker cp "$$container_id":/output/crystal-$(CRYSTAL_VERSION)-$(PACKAGE_ITERATION).tar $@ \
 	  && docker rm -v "$$container_id"
 
 .PHONY: compress64
 compress64: $(OUTPUT_BASENAME64).tar.gz $(OUTPUT_BASENAME64).tar.xz ## Build compressed tarballs
 
-$(OUTPUT_BASENAME64).tar.gz: $(OUTPUT_BASENAME64).tar
-	gzip -c $(OUTPUT_BASENAME64).tar > $(OUTPUT_BASENAME64).tar.gz
+$(OUTPUT_BASENAME64).tar.gz $(OUTPUT_BASENAME32).tar.gz: %.gz: %
```
straight-shoota: Are the first two targets actually necessary?

Daniel-Worrall: Not necessary, but it restricts it to a static readable list. It can be removed in the interest of brevity but I kept them in in the interest of clarity.

straight-shoota:
The first commit of this PR has been merged in #139. The others don't really apply anymore due to the removal of distro package building recipes.
I think this can be closed.

PR #52, opened 2020-01-15
Add git shallow clones

Daniel-Worrall:
Optimises git cloning with shallow clone.

Reduces processing time & file size.


RX14:
Unfortunately this doesn't allow checking out a specific sha1 revision, unless it *happens* to be tagged.

If you ask for a specific revision from github it will refuse, unless the revision corresponds to an existing ref, such as a tag or a branch.

straight-shoota:
I guess then there's nothing to do here.

Daniel-Worrall:
I can remove the changes with references to the crystal sha and keep the ones that use a tagged version like bdwgc, libatomic and shards. If it changes in the future that we need to use a specific sha, we could use a conditional.

RX14:
@Daniel-Worrall there's scope for wanting to build with a custom GC version in the future but I think that would be fine

Daniel-Worrall:
Done.

Also, I noticed that there could be further optimizations such as in the Linux Dockerfile, various things are cloned twice. These could be copied over to the 2nd build stage before altered so that it saves on cloning. The only problem with this is that currently all the cloning/configuration is done on a single RUN and splitting that up might just end up removing any optimizations. 
Anyway, I'm not going to delve into it any further.

PR #53, opened 2020-01-21
Fix APK packages in alpine.Dockerfile

straight-shoota:
* Remove `readline-dev` because `Readline` has been removed from stdlib since Crystal 0.32.0
* Use proper packages providing `libevent`, `openssl` and `zlib` static libraries. (`*-static` packages also install `*-dev` packages

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -3,9 +3,9 @@ FROM alpine:3.11 as runtime
 RUN \
   apk add --update --no-cache --force-overwrite \
     # core dependencies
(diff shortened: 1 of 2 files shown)


bcardiff:
Could you remove libreadline-dev from ubuntu.Dockerfile also?

I am not sure the zlib-static will be picked since pkg-config is not available and `@[Link("z")]` is used for `LibZ`. But as long as zlib-dev is included it should be fine to start with.

straight-shoota:
Done.

bcardiff:
The PR in the main repo can be updated again now. Thanks!

PR #55, opened 2020-01-28
Fix apk static packages in alpine.Dockerfile

straight-shoota:
Use proper packages providing libevent, openssl and zlib static libraries.
The *-static packages usually also install *-dev packages (except openssl-libs-static).

Followup to #29


straight-shoota:
A maintenance build with this change is running at https://circleci.com/workflow-run/8ef88421-451d-4be6-8c29-8986b7390b65

PR #63, opened 2020-04-28
Build libgc in alpine image to enable MT builds

Blacksmoke16:
~One question I have left is what to do about `LIBRARY_PATH`?  Given I think both paths need to be used?~

~Added built libgc path to `LIBRARY_PATH` in build image.~

`LIBRARY_PATH` is no longer required apparently due to #69.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -18,8 +18,8 @@ LIBATOMIC_OPS_VERSION = v7.6.10
 
 OUTPUT_DIR := build
 BUILD_CONTEXT := $(CURDIR)/build-context
-BUILD_ARGS_UBUNTU64 := --build-arg crystal_deb=crystal.deb $(BUILD_CONTEXT)/ubuntu64 --build-arg base_docker_image=ubuntu:bionic --build-arg gc_version=$(GC_VERSION) --build-arg libatomic_ops_version=$(LIBATOMIC_OPS_VERSION)
-BUILD_ARGS_UBUNTU32 := --build-arg crystal_deb=crystal.deb $(BUILD_CONTEXT)/ubuntu32 --build-arg base_docker_image=i386/ubuntu:bionic --build-arg gc_version=$(GC_VERSION) --build-arg libatomic_ops_version=$(LIBATOMIC_OPS_VERSION)
+BUILD_ARGS_UBUNTU64 := --build-arg crystal_deb=crystal.deb $(BUILD_CONTEXT)/ubuntu64 --build-arg base_docker_image=ubuntu:focal
+BUILD_ARGS_UBUNTU32 := --build-arg crystal_deb=crystal.deb $(BUILD_CONTEXT)/ubuntu32 --build-arg base_docker_image=i386/ubuntu:bionic
 BUILD_ARGS_ALPINE := --build-arg crystal_targz=crystal.tar.gz $(BUILD_CONTEXT)/alpine --build-arg gc_version=$(GC_VERSION) --build-arg libatomic_ops_version=$(LIBATOMIC_OPS_VERSION)
 DOCKER_TAG_UBUNTU := $(DOCKER_REPOSITORY):$(DOCKER_TAG)
 DOCKER_TAG_ALPINE := $(DOCKER_REPOSITORY):$(DOCKER_TAG)-alpine
@@ -48,17 +48,13 @@ alpine: ## Build alpine images
 alpine: $(OUTPUT_DIR)/docker-$(CRYSTAL_VERSION)-alpine.tar.gz
 alpine: $(OUTPUT_DIR)/docker-$(CRYSTAL_VERSION)-alpine-build.tar.gz
 
-$(BUILD_CONTEXT)/ubuntu32: $(BUILD_CONTEXT)/ubuntu32/crystal.deb
-	cp ubuntu.Dockerfile $@/Dockerfile
-	mkdir $@/files/
-	cp ../linux/files/feature-thread-stackbottom-upstream.patch $@/files/feature-thread-stackbottom-upstream.patch
+$(BUILD_CONTEXT)/ubuntu32: ubuntu32.Dockerfile $(BUILD_CONTEXT)/ubuntu32/crystal.deb
+	cp ubuntu32.Dockerfile $@/Dockerfile
 
-$(BUILD_CONTEXT)/ubuntu64: $(BUILD_CONTEXT)/ubuntu64/crystal.deb
+$(BUILD_CONTEXT)/ubuntu64: ubuntu.Dockerfile $(BUILD_CONTEXT)/ubuntu64/crystal.deb
 	cp ubuntu.Dockerfile $@/Dockerfile
-	mkdir $@/files/
-	cp ../linux/files/feature-thread-stackbottom-upstream.patch $@/files/feature-thread-stackbottom-upstream.patch
 
-$(BUILD_CONTEXT)/alpine: $(BUILD_CONTEXT)/alpine/crystal.tar.gz
+$(BUILD_CONTEXT)/alpine: alpine.Dockerfile $(BUILD_CONTEXT)/alpine/crystal.tar.gz
 	cp alpine.Dockerfile $@/Dockerfile
 	mkdir $@/files/
 	cp ../linux/files/feature-thread-stackbottom-upstream.patch $@/files/feature-thread-stackbottom-upstream.patch
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -13,13 +13,13 @@ CRYSTAL_TARGZ ?=     ## Which crystal.tar.gz file to install in docker images (a
 DOCKER_TAG ?=	     ## How to tag the docker image (examples: `0.27.2`, `nightly20190307`). `-build` will be appended for build images.
 DOCKER_REPOSITORY ?= ## Docker hub repository to commit image
 
-GC_VERSION = b26e281bcac33444f523862b97f81e2a74bc2b4d
+GC_VERSION = v8.0.4
 LIBATOMIC_OPS_VERSION = v7.6.10
 
 OUTPUT_DIR := build
 BUILD_CONTEXT := $(CURDIR)/build-context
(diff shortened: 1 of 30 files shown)


Blacksmoke16:
Bump.

Blacksmoke16:
Would be great if this made it in time for https://github.com/crystal-lang/crystal/pull/9317 in order to enable MT builds with `0.35.0`.

\cc @bcardiff @RX14 @waj

docker/alpine.Dockerfile
```diff
@@ -3,11 +3,28 @@ FROM alpine:3.11 as runtime
...
+ARG libatomic_ops_version
+RUN git clone https://github.com/ivmai/bdwgc \
+ && cd bdwgc \
+ && git checkout ${gc_version} \
+ && git clone https://github.com/ivmai/libatomic_ops \
+ && (cd libatomic_ops && git checkout ${libatomic_ops_version}) \
+ \
+ && ./autogen.sh \
+ && ./configure --disable-debug --disable-shared --enable-large-config \
+ && make -j$(nproc) CFLAGS=-DNO_GETCONTEXT
```
RX14, line 25: It'll be much better to just `make install` here than edit `LIBRARY_PATH`.

Blacksmoke16: Would still need to manually add it to the other one, otherwise it would be overridden I assume?

docker/Makefile
```diff
@@ -13,11 +13,14 @@ CRYSTAL_TARGZ ?=     ## Which crystal.tar.gz file to install in docker images (a
 DOCKER_TAG ?=	     ## How to tag the docker image (examples: `0.27.2`, `nightly20190307`). `-build` will be appended for build images.
 DOCKER_REPOSITORY ?= ## Docker hub repository to commit image
 
+GC_VERSION = b26e281bcac33444f523862b97f81e2a74bc2b4d
```
bcardiff, line 16: I think we should use the same organization as in linux directory. Doing the patch with `files/feature-thread-stackbottom-upstream.patch` on top of v8.0.4

docker/alpine.Dockerfile
RX14: You copied from the dockerfile which builds the release tarballs. That container only exists to produce a single tar.gz, so you can take shortcuts. This container is meant to be used, so it's best to do it directly.

And how about doing this in the ubuntu container too?

docker/Makefile
Blacksmoke16: Hmm, I tried doing this but it seems applying the patch fails?

```
Step 5/11 : COPY files/feature-thread-stackbottom-upstream.patch /tmp/
 ---> 0f1be76db0e2
Step 6/11 : RUN git clone https://github.com/ivmai/bdwgc  && cd bdwgc  && git checkout ${gc_version}  && git clone https://github.com/ivmai/libatomic_ops  && (cd libatomic_ops && git checkout ${libatomic_ops_version})   && patch -p1 < /tmp/feature-thread-stackbottom-upstream.patch   && ./autogen.sh  && ./configure --disable-debug --disable-shared --enable-large-config  && make -j$(nproc) CFLAGS=-DNO_GETCONTEXT  && make install
 ---> Running in bb241fa8b495
Cloning into 'bdwgc'...
Note: switching to 'v8.0.4'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at d3dede3c [8.0.4]
Cloning into 'libatomic_ops'...
Note: switching to 'v7.6.10'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 7702826 [7.6.10]
patching file include/gc.h
Hunk 1 FAILED 2200/2200.
     GC_blocked_sp = NULL;
 }
 
+  GC_API void GC_CALL GC_set_stackbottom(void *gc_thread_handle,
+                                         const struct GC_stack_base *sb)
+  {
+    GC_ASSERT(sb -> mem_base != NULL);
+    GC_ASSERT(NULL == gc_thread_handle || &GC_stackbottom == gc_thread_handle);
+    GC_ASSERT(NULL == GC_blocked_sp
+              && NULL == GC_traced_stack_sect); /* for now */
+    (void)gc_thread_handle;
+
+    GC_stackbottom = (char *)sb->mem_base;
+#   ifdef IA64
+      GC_register_stackbottom = (ptr_t)sb->reg_base;
+#   endif
+  }
+
+  GC_API void * GC_CALL GC_get_my_stackbottom(struct GC_stack_base *sb)
+  {
+    GC_ASSERT(GC_is_initialized);
+    sb -> mem_base = GC_stackbottom;
+#   ifdef IA64
+      sb -> reg_base = GC_register_stackbottom;
+#   endif
+    return &GC_stackbottom; /* gc_thread_handle */
+  }
 #endif /* !THREADS */
 
 GC_API void * GC_CALL GC_do_blocking(GC_fn_type fn, void * client_data)
Hunk 3 FAILED 913/920.
   /* Adjust our stack bottom pointer (this could happen unless  */
   /* GC_get_stack_base() was used which returned GC_SUCCESS).   */
   GC_ASSERT(me -> stack_base != NULL);
-  if ((word)me->stack_base < (word)(&stacksect))
+  if ((word)me->stack_base < (word)(&stacksect)) {
     me -> stack_base = (ptr_t)(&stacksect);
+#   if defined(I386)
+      me -> initial_stack_base = me -> stack_base;
+#   endif
+  }
 
   if (me -> thread_blocked_sp == NULL) {
     /* We are not inside GC_do_blocking() - do nothing more.    */
patching file misc.c
patching file pthread_support.c
patching file tests/test.c
patching file win32_threads.c
The command '/bin/sh -c git clone https://github.com/ivmai/bdwgc  && cd bdwgc  && git checkout ${gc_version}  && git clone https://github.com/ivmai/libatomic_ops  && (cd libatomic_ops && git checkout ${libatomic_ops_version})   && patch -p1 < /tmp/feature-thread-stackbottom-upstream.patch   && ./autogen.sh  && ./configure --disable-debug --disable-shared --enable-large-config  && make -j$(nproc) CFLAGS=-DNO_GETCONTEXT  && make install' returned a non-zero code: 1
make: *** [Makefile:97: build/docker-0.35.0-dev-alpine.tar.gz] Error 1
make: Leaving directory '/home/george/dev/git/distribution-scripts/docker'
```

I'm probably missing something.  As far as I can tell this is exactly what the one in the linux imagine is doing, minus the `make install`.

Blacksmoke16: This ended up being just an old `patch` binary.  Updating it fixed it.

docker/alpine.Dockerfile
Blacksmoke16: Done

Blacksmoke16:
@RX14 @bcardiff @straight-shoota This should be good for another review.

straight-shoota:
I suppose the build tools should be removed from the docker image after building libgc?

tducasse:
Hey just FYI, I was looking for an "official" alpine crystal image and stumbled onto this PR.

I was trying to build a static binary for Crystalline to solve [this issue](https://github.com/elbywan/crystalline/issues/7). It seems like `yaml-dev` on Alpine doesn't give the right static libs, so I had to add `yaml-static` as well.

Not sure if it is useful, but since this is not merged yet, I thought I'd share 😄

caspiano:
Bump :)

straight-shoota:
This needs conflict resolution.

I'm also a bit confused now about the Ubuntu dockerfile changes. Description only talks about alpine. And a patched library should be shipped with the debian package, so rebuilding seems unnecessary here.

caspiano:
@Blacksmoke16 Made a PR to your repo that I believe resolves the comment above by @straight-shoota

Blacksmoke16/distribution-scripts/pull/1

Blacksmoke16:
@caspiano Thanks!  Merged and fixed up two things I noticed.  Should be good for another review now.

Disclaimer:  I didn't test it locally yet.

docker/alpine.Dockerfile
```diff
@@ -3,11 +3,33 @@ FROM alpine:3.13 as runtime
...
+ \
+ && patch -p1 < /tmp/feature-thread-stackbottom-upstream.patch \
+ \
+ && ./autogen.sh \
+ && ./configure --disable-debug --disable-shared --enable-large-config \
+ && make -j$(nproc) CFLAGS=-DNO_GETCONTEXT \
+ && make install
+
+# Remove build tools from image now that libgc is built
+RUN apk del -r --purge autoconf automake libtool patch
```
Blacksmoke16, line 32: Actually do we want to uninstall `patch`?  I'm remembering that it was installed by default but the version was too old or something so had to update it.

straight-shoota: If it was installed before, then let's just leave it.

Maybe it's not even necessary to install it explicitly, though?

Blacksmoke16: Possibly, given the image is newer than it was before.  I'll try it out later today.

Blacksmoke16: Ok, I think we're good the way it is.  The base alpine image doesn't have `patch` by default, so it makes sense to uninstall it when we're done.

straight-shoota (approved):

bcardiff (approved):

caspiano:
This would be awesome to have with `1.1.0` 😉 What's currently in the way of a merge?
I cannot wait to run my crystal services with MT 🤤

straight-shoota:
> I cannot wait to run my crystal services with MT 🤤

Well, you can already do that with older releases, just need to provide the patched library.

caspiano:
I normally run alpine images, I'd either have to build an image with the patch (like this) or switch to the ubuntu images.
This is a great change, thanks @Blacksmoke16!

PR #71, opened 2020-06-19
Avoid requiring shards to build shards

bcardiff:
This changes uses `make lib` so it does not require shards to be available.

This is also needed to unlock builds of 1.0.0-dev (since molinillo states `crystal: 0.x`)

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -94,7 +94,7 @@ ARG musl_target
 RUN git clone https://github.com/crystal-lang/shards \
  && cd shards \
  && git checkout ${shards_version} \
- && shards install --production \
+ && make lib \
  \
  && /crystal/bin/crystal build --stats --target ${musl_target} \
     ./src/shards.cr -o shards --static ${release:+--release} \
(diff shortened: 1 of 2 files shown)


linux/Dockerfile
```diff
@@ -94,7 +94,7 @@ ARG musl_target
 RUN git clone https://github.com/crystal-lang/shards \
  && cd shards \
  && git checkout ${shards_version} \
- && shards install --production \
+ && make lib \
  \
  && /crystal/bin/crystal build --stats --target ${musl_target} \
     ./src/shards.cr -o shards --static ${release:+--release} \
```
straight-shoota, line 100: We could also consider to use `make` for building the binary. 
```suggestion
 && CRYSTAL_OPTS="--stats --target ${musl_target} --static ${release:+--release}" make \
```

jhass: That  reminds me we should give the shards Makefile  a `release=1` option like we got for the crystal one.

bcardiff: I want to unify a bit the makefile arguments of shards and crystal a bit. Like in https://github.com/crystal-lang/shards/compare/master...bcardiff:feature/makefile-flags

That will need to wait for a shard release before it is used here.

Mind if we update first that and then we unify how shards is build across the scripts?

PR #72, opened 2020-07-30
Bintray publishing scripts

bcardiff:
This PR handles the publication of deb and rpm packages to [Bintray](https://bintray.com/crystal).

The process to transform linux 64 and 32 bits .tar.gz to .deb and .rpm packages via fpm is duplicated from the `./linux` directory. Eventually the one in `./linux` will go away narrowing it's resposibility to produce the .tar.gz only.

The bintray packages are published in three channels: stable, unstable, nightly. There are helper scripts in `./bintray/publish-<channel>.sh` to download .tar.gz, build deb/rpm and publish them directly to bintray.

The deb repository will be registered as

```
deb https://dl.bintray.com/crystal/deb all stable
deb https://dl.bintray.com/crystal/deb all unstable
deb https://dl.bintray.com/crystal/deb all nightly
```

There are no distro distinctions needed.

The rpm repository will be registered as

```
[crystal]
name=Crystal
baseurl=https://dl.bintray.com/crystal/rpm/{DISTRO}/x86_64/{CHANNEL}
gpgcheck=0
repo_gpgcheck=1
gpgkey=http://bintray.com/user/downloadSubjectPublicKey?username=bintray
```

where `DISTRO=el6 | all` and `CHANNEL=stable | unstable | nightly` to expose properly dependencies to CentOS 6/7/8 and Fedora. Only CentOS 6 is the one that needs a different expression of dependencies.

There are some scripts I used to test the packages in different distros via docker images: `./bintray/test-install.sh`

The packages are uploaded signed with the same distribution key as before, but the metadata of the repository will be handled by bintray itself. All these settings are documented and can be recreated by the Makefile.

To handle the signing in an automated way the `./bintray/sign.sh` will create some docker containers with the crystal distribution signing key. This is based on the existing work of the crystal-dist repo which will eventually stop been used to publish packages.

After this is merged the crystal repo will be updated to publish nightly packages to these repositories.

The `./bintray/scripts/install.sh` script will help users to do fresh installs and upgrades to specific crystal version packages. This script will be advertised with a nicer url but should be kept here since it is tested with the `test-install.sh` script already mentioned.


bintray/scripts/install.sh
```diff
@@ -0,0 +1,157 @@
...
+  [[ -z "$DISTRO_TYPE" ]] || return
+
+  [[ -x "/usr/bin/apt-get" ]] && DISTRO_TYPE="deb" && return
+  [[ -x "/usr/bin/yum" ]]     && DISTRO_TYPE="rpm" && return
+
+  if [[ -z "$DISTRO_TYPE" ]]; then
+    _error "Unable to identify distribution. Please, report to https://forum.crystal-lang.org/c/help-support/11"
+    exit 1
+  fi
+}
```
j8r, line 73: It's simpler, and more reliable, to use `ID` vairable in `/etc/os-release`, and use `case ... in`. It is resent starting CentOS 7, and all Debians (I tested on 7).

You can just import the file, and the variables will be set.

For the `deb`/`rpm`, `hash yum 2>/dev/null` and `hash apt-get 2>/dev/null` can be used, because at the end we want to know if this binaries are present in the `PATH`, not specifically in `/usr/bin`

bintray/sign.sh
```diff
@@ -0,0 +1,83 @@
...
+# ./sign.sh import-key-deb
+# ./sign.sh import-key-rpm
+# ./sign.sh sign-deb build/unsigned/crystal_[VERSION]-[ITERATION]_[ARCH].deb
+#   # Output at build/signed/crystal_[VERSION]-[ITERATION]_[ARCH].deb
+# ./sign.sh sign-rpm build/unsigned/crystal-[VERSION]-[ITERATION].[distro].[ARCH].rpm
+#   # Output at build/signed/crystal-[VERSION]-[ITERATION].[distro].[ARCH].rpm
+# ./sign.sh clean
+# ```
+
+function debian() {
```
j8r, line 23:
```suggestion
debian() {
```

bintray/sign.sh
```diff
@@ -0,0 +1,83 @@
...
+  docker-compose exec debian /bin/sh -c "$@"
+}
+
+function centos() {
```
j8r, line 27:
```suggestion
centos() {
```

bintray/rebuild.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+make create_bintray_deb_repo create_bintray_rpm_repo
+
+function rebuild () {
```
j8r, line 5:
```suggestion
rebuild () {
```

bintray/scripts/install.sh
j8r: For CentOS 6, `/etc/issue` can be used. instead (fallback if `/etc/os-release` is not present)

j8r: Antother option is not to support it the EOL is 2020-11-30.

j8r: Another option is not to support it the EOL is 2020-11-30.

bcardiff: We support CentOS 6. I discuss pros and cons of loading /etc/os-release and compare different code base that does a similar task and decided to base the work on [icy/pacapt](https://github.com/icy/pacapt/blob/ng/pacapt#L244-L302).

Not all distros has os-release and this script worked fine on all the test environments.

For the deb/rpm, I agree that using hash or which would be better to avoid them beeing in /usr/bin. Yet, is that a reasonable scenario?

j8r: The `hash rpm`/`hash apt-get` is enough, because I don't see any logic related to if the distribution is Debian vs Ubuntu, or Fedora vs CentOS.

That's not really if there is a scenario, that just makes sense to test if `apt-get` exists if we want to use it afterwards.
If `/usr/bin/apt-get` is checked, logically `/usr/bin/apt-get` has to be used.

I find this function too complex for one simple thing: know what is the package manager, rpm, deb, or error?

Also, `PACKAGE_MANAGER` would be more correct than `DISTRO_TYPE` IMHO. For me `DIST*`, like in `/etc/lsb-release`, is related to the distribution family, not the package manager.

j8r: i.e.:
```sh
if hash apt-get 2>/dev/null ;then
  PACKAGE_MANAGER="deb"
elif hash yum 2>/dev/null ;then
  PACKAGE_MANAGER="rpm"
else
  _error "..."
  exit 1
fi
```

bcardiff: I've just applied some suggestions. Thanks.

j8r: I see, no problem. Hum, you kept `_match_etc_issue`? It is no longer needed (unless you want to know the distribution in the future?)

bintray/scripts/install.sh
```diff
@@ -0,0 +1,158 @@
...
+    DISTRO_TYPE="deb"
+  elif hash yum 2>/dev/null; then
+    DISTRO_TYPE="rpm"
+  else
+    _error "Unable to identify distribution. Please, report to https://forum.crystal-lang.org/c/help-support/11"
+    exit 1
+  fi
+}
+
+set -eu
```
j8r, line 76: To remove, already declared at line 37 (by the way, it is usually done just have the shebang)

bintray/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile-sign-debian
+    volumes:
+      - gnupg_debian:/root/.gnupg
+      - ./build:/build
+
+volumes:
+  gnupg_centos:
+  gnupg_debian:
```
j8r, line 23: You're sure to sure to introduce a new requirement to use distribution scripts? `docker-compose` was not used anywhere in the repo before.
Being very simple, this dependency can be easily avoided?

I'm thinking about other people using it. However it's unlikely I use all of this myself, so that's your decision :shrug:

bcardiff: It's a migration from https://github.com/crystal-lang/crystal-dist where docker-compose was already used. 

This is more public infrastructures script in case someone find it useful rather than something expected to be used by somebody else.

PR #79, opened 2021-01-27
Fix alpine Dockerfile to add *-static packages and update to alpine 3.13

straight-shoota:
Resolves crystal-lang/crystal#10313

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -5,7 +5,7 @@ RUN \
     # core dependencies
     gc-dev gcc gmp-dev libatomic_ops libevent-static musl-dev pcre-dev \
     # stdlib dependencies
-    libxml2-dev openssl-dev openssl-libs-static tzdata yaml-dev zlib-static \
+    libxml2-dev openssl-libs-static tzdata yaml-static zlib-static \
     # dev tools
     make git
 
@@ -31,6 +31,6 @@ FROM runtime as build
 
 RUN \
   apk add --update --no-cache --force-overwrite \
-    llvm10-dev llvm10-static g++
+    llvm10-static g++
 
 CMD ["/bin/sh"]
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.12 as runtime
+FROM alpine:3.13 as runtime
 
 RUN \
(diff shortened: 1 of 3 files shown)


bcardiff:
In https://github.com/crystal-lang/distribution-scripts/pull/68 that was changed here https://github.com/crystal-lang/distribution-scripts/pull/68/files#diff-c306b13048e7d0a5c50f527d9ff3fbcd7deb85a451c5bdad725a10f9e2b2322cL48 but not here https://github.com/crystal-lang/distribution-scripts/blob/master/docker/alpine.Dockerfile#L8

Are you sure you prefer to use different alpine base images while building the binaries and distributing the docker images?

If we want to bump entirely to alpine 3.13 we need to change `alpine_image=alpine:3.12` and `alpine_image=i386/alpine:3.12` also.

Whatever is decided, I can rebuilt the alpine 0.36 image with these changes and republish it.

straight-shoota:
> In https://github.com/crystal-lang/distribution-scripts/pull/68 that was changed here https://github.com/crystal-lang/distribution-scripts/pull/68/files#diff-c306b13048e7d0a5c50f527d9ff3fbcd7deb85a451c5bdad725a10f9e2b2322cL48 but not here https://github.com/crystal-lang/distribution-scripts/blob/master/docker/alpine.Dockerfile#L8

Yeah, I noticed that, too. We should probably try to avoid this duplication. IIRC there was some effort about unifying both dockerfiles / re-using the alpine dockerfile for the linux build. Would've definitely helped here.

> Are you sure you prefer to use different alpine base images while building the binaries and distributing the docker images?

I didn't look into the base image version used for linux build. But I suppose we should probably update that to 3.13 as well.

bcardiff (approved):
Should I re-upload the 0.36 alpine docker image with alpine:3.13 or 3.12 then?

straight-shoota:
3.13 sounds better I think.

docker/alpine.Dockerfile
```diff
@@ -31,6 +31,6 @@ FROM runtime as build
 
 RUN \
   apk add --update --no-cache --force-overwrite \
-    llvm10-dev llvm10-static g++
```
bcardiff, line 34: I think it is needed https://github.com/crystal-lang/crystal/pull/10318/checks?check_run_id=1779776162#step:5:43

PR #80, opened 2021-01-27
Restore llvm10-dev in alpine images

bcardiff:
Ref: https://github.com/crystal-lang/distribution-scripts/pull/79/files#r565621160


straight-shoota (approved):
`llvm10-dev` should be a subpackage of `llvm10-static` 🤷 Maybe I'm missing something, though.

bcardiff:
Without it the llvm-config is not available 🤷

PR #83, opened 2021-03-10
Update shards to 0.14.1

bcardiff:
This updates shards to 0.14.1

Sample workflow https://app.circleci.com/pipelines/github/crystal-lang/crystal/5428/workflows/6f96f56f-c429-4941-b352-2d2fb4aff1c0

It is based on the commit before #79 because after that PT the crystal binaries end up with some shared libraries linked and the maintenance_release complains https://app.circleci.com/pipelines/github/crystal-lang/crystal/5427/workflows/8f62a04e-4e79-4633-8952-77e7c777dc00 .

PR #87, opened 2021-04-12
Remove compilation of libcrystal.a

maxfierke:
With https://github.com/crystal-lang/crystal/pull/10463 being merged, `libcrystal.a` no longer exists

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -15,15 +15,8 @@ COPY ${crystal_targz} /tmp/crystal.tar.gz
 RUN \
   tar -xz -C /usr --strip-component=1  -f /tmp/crystal.tar.gz \
     --exclude */lib/crystal/lib \
-    --exclude */share/crystal/src/llvm/ext/llvm_ext.o \
-    --exclude */share/crystal/src/ext/libcrystal.a && \
+    --exclude */share/crystal/src/llvm/ext/llvm_ext.o && \
   rm /tmp/crystal.tar.gz
-
-# Build libcrystal
-RUN \
-  cd /usr/share/crystal && \
-  cc -fPIC -c -o src/ext/sigfault.o src/ext/sigfault.c && \
-  ar -rcs src/ext/libcrystal.a src/ext/sigfault.o
 
 CMD ["/bin/sh"]
 
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -27,15 +27,6 @@ RUN git clone https://github.com/ivmai/bdwgc \
  && ./configure --disable-debug --disable-shared --enable-large-config \
  && make -j$(nproc)
 
-# Cross-compile crystal and build libcrystal.a
-ARG crystal_sha1
-ARG musl_target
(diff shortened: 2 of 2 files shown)


straight-shoota (approved):

straight-shoota:
There is also a mention of libcrystal.a in `docker/alpine.Dockerfile` left.

straight-shoota:
Running maintenance_release workflow in https://app.circleci.com/pipelines/github/crystal-lang/crystal/5591/workflows/33c792de-0047-4e84-aa30-5c816d174c6e

straight-shoota:
The maintenance build failed (https://app.circleci.com/pipelines/github/crystal-lang/crystal/5591/workflows/33c792de-0047-4e84-aa30-5c816d174c6e/jobs/59375).

The failure is caused by `[ "$(ldd .build/crystal 2&> i | wc -l)" -eq "1" ]`. It breaks because `ldd .build/crystal` doesn't print a line to stdout anymore. It just prints to stderr.

Somehow the removal of `libcrystal.a` has changed the behaviour of Alpine's ldd.
```shell
$ ldd /usr/lib/crystal/bin/crystal
        /lib/ld-musl-x86_64.so.1 (0x7f41d908b000)
$ ldd .build/crystal
/lib/ld-musl-x86_64.so.1: .build/crystal: Not a valid dynamic program
$ file /usr/lib/crystal/bin/crystal
/usr/lib/crystal/bin/crystal: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), statically linked, with debug_info, not stripped
$ file .build/crystal
.build/crystal: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, with debug_info, not stripped
```

I don't know why this change appears, but it seems we'll have to change the static executable check.
Changing the expected line count to `0` would probably be fine. Although that would probably also succeed with any other kind of error. So maybe checking `file build/crystal | grep statically linked` would be better? `file` is an additional dependency, though.

On a similar though, it might also be a good idea to test executing the produced binary. At least `crystal help` should serve as a smoke test that it is indeed runnable.

bcardiff:
@straight-shoota The issue something with #79. That is way 1.0.0 is built with that PR excluded as stated in https://github.com/crystal-lang/distribution-scripts/pull/83#issue-589836036

straight-shoota:
Okay, so it's unrelated to this. 

The CI failure linked from #83 shows exactly the same behaviour. But `Not a valid dynamic program` seems very explicit that it does not link some shared libraries. I presume it's just that the linker changed the output type. But that doesn't really matter. The only difference between `LSB shared object` and `LSB executable` seems to be that the former could be linked against as a dynamic library. That's irrelevant for our use case. The binaries are statically linked in both cases.

So I think we can just fix the static linking check on top of #79/master, and then rebase this PR.

straight-shoota:
The maintenance build after #88 was merged succeeds: https://app.circleci.com/pipelines/github/crystal-lang/crystal/5600/workflows/fcf1003b-9115-41bf-a546-b2c2643ab3d1

PR #88, opened 2021-04-12
Fix static executable checks

straight-shoota:
Fixes static executable checks which currently fails on alpine 3.13 (see https://github.com/crystal-lang/distribution-scripts/pull/83#issue-589836036).

The reason is explained in https://github.com/crystal-lang/distribution-scripts/pull/87#issuecomment-817865398

Follow-up to #79


j8r:
I'll nitpick, a good practice is to use `=` instead of `-eq`. Plus, in my opinion, it is easier to understand because closer to other languages.

straight-shoota:
`-eq` is integer comparison, `=` is string comparison. There's no practical difference in this case, but `-eq` is formally more correct.

bcardiff (approved):

PR #91, opened 2021-04-29
Update installer script

straight-shoota:
https://github.com/crystal-lang/crystal/issues/10626


straight-shoota:
Git doesn't pick up that `bintray/scripts/install.sh` was moved to `packages/scripts/install.sh` and modified (probably too many modifications).

The script changes a lot because OBS has specific repos for each distribution release.
So instead of checking for deb vs. rpm, we need more specific information. This is drawn from `/etc/os-release` which should usually be available. If it does not work, the env variable `DISTRO_REPO` can be used to specify a value.

Alternative channels are still missing, mostly because it's hard to set up on OBS without current nightly releases working. I'll add that once our nightlies are back.

I've added test cases for more relevant distro releases. 

There were a couple minor issues with i386 packages themselves (ATM rebuilding and re-checking), but the installer should be good.

beta-ziliani (approved):
Looks good to me

bcardiff:
Looks good. Thanks for keeping the tests.

- Are the packages for 0.x going to be republished as is or rebuilt to be available on obs?
- How is the integration with nightlies and tagged release going to work? As in, how we can trigger the build in obs?

straight-shoota:
Yeah, the tests are great 👍 

* I haven't looked into providing 0.x releases yet. I suppose it should be relatively easy to push them through OBS. Not sure if it's much useful, though. By now most people should be using 1.0 (hopefully).
* Tagged releases should be working with a simple webhook. OBS then pulls in the tarball from the github release automatically. For nightlies we probably need to actively push the tarball from CI (or provide them somewhere for a pull), using the `osc` client.

bcardiff:
So OBS receives a tarball and package it as .deb/.rpm.
I think that whatever is needed to push a tarball should be included in this repo also and then used in the main repo. So we can have nightlies in OBS.

Another task artifact that would need to be updated at some point is the travis build script at https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/crystal.rb#L104-L106 .

bcardiff (approved):

PR #93, opened 2021-05-03
Fix packages test script to use deb noninteractive

straight-shoota:


bcardiff (approved):
The tzinfo package requesting for location, right?

straight-shoota:
Yep.

PR #96, opened 2021-05-13
Update omnibus and build with LLVM 10 on macOS 

### LINKED ISSUE #89: Bumping LLVM version on macOS?
My understanding from this line: https://github.com/crystal-lang/distribution-scripts/blob/e5f21df883d800e9e68f63b79cf796696f905cf1/omnibus/config/software/llvm.rb#L2 is that LLVM for the binary distribution uses LLVM 6. Is there a particular reason for holding this back on LLVM 6?

Would it be possible to bump this to something much newer, perhaps LLVM 10 (or ideally LLVM 11+)?

The reason I ask is that I believe either LLVM 10 or 11 is needed to target AArch64 (ARM64) on Darwin/macOS. If were to use a newer LLVM and also set `-DLLVM_TARGETS_TO_BUILD=X86;AArch64` for the builds of LLVM used to build the distribution releases, we could use the official x86_64 binary release to bootstrap a ARM64 Homebrew build

maxfierke:
Picking up where @bcardiff was working last August on building with LLVM 10. I think this will get us closer to being able to build for both x86_64 and arm64 for macOS, but will be working on that more separately.

Fixes #89 

Need to do more testing with it, but opening a PR for feedback


bcardiff:
Hi @maxfierke thanks for investing energy on this. I have a couple of questions to understand why some of these changes were needed or wanted on your side.

I think the goal should be to have a build backed by llvm 10 (any 10 version, I recall there were some issues with llvm 10.0.1 and [how libxml was advertised](https://discourse.brew.sh/t/llvm-config-10-0-1-advertise-libxml2-tbd-as-system-libs/8593))

- I see you are setting `LLVM_ENABLE_LIBXML2=OFF`. Do you remember what pushed you to set that? Was the library linked somehow on the llvm 10 output? Or something later failed because of it. Again this might be related to the linked topic.
- Bumping dependencies in general is good, but was there other motivations to bump omnibus and openssl. This triggers other updates I am not sure how to review.

maxfierke:
> * I see you are setting `LLVM_ENABLE_LIBXML2=OFF`. Do you remember what pushed you to set that? Was the library linked somehow on the llvm 10 output? Or something later failed because of it. Again this might be related to the linked topic.

Yeah, there were some linking issues with `libxml2.tbd`, if I remember correctly. The linked topic looks related.

> * Bumping dependencies in general is good, but was there other motivations to bump omnibus and openssl. This triggers other updates I am not sure how to review.

Omnibus was bumped for getting it running on macOS Big Sur. Wasn't having any luck with the prior version. With that bump, I removed or bumped some of the pieces that overlap in Omnibus, specifically OpenSSL and removing the older copy of cacerts, because it's good to keep that stuff as up to date as possible.

PR #98, opened 2021-06-05
Update omnibus

maxfierke:
This PR updates omnibus to a version with support for running and compiling different packages on an ARM64-based Mac (i.e. the M1)

Through this, I introduced `-DLLVM_ENABLE_LIBXML2=OFF` (discussed also in #96) to the LLVM build, as well. On Big Sur, when compiling LLVM 10, you end up with a garbage `-llibxml2.tdd` in the default system libs. It doesn't seem to be needed. This is fixed in LLVM 11, but we're not there yet.

This is the first PR for compiling the Crystal omnibus package for ARM64 on LLVM 10.


omnibus/.ruby-version
```diff
@@ -1 +1 @@
-2.5.6
+2.7.3
```
bcardiff, line 1: Note: we will need to sync `.circleci/config.yml` in `dist_darwin` on the crystal-lang/crystal repo

```
ruby-install ruby 2.5.6
```

and

```
echo "2.5.6" > /tmp/workspace/distribution-scripts/.ruby-version
```

omnibus/Gemfile
```diff
@@ -5,5 +5,5 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-gem 'omnibus', '~> 6.1.4'
-gem 'omnibus-software', github: 'chef/omnibus-software', ref: 'ae2a2f1cae3e8d80197a1cac5311b910300a53ff'
+gem 'omnibus', github: 'chef/omnibus', ref: '1d97cd9e79a062dfc12f0c047215f3a9e4d65b09'
```
bcardiff, line 8: Note: This commit is included in 8.1.10 8.1.9 8.1.8 tags, but those are not published in rubygems for some reason. ✅

omnibus/config/projects/llvm.rb
```diff
@@ -18,7 +18,7 @@ def valid_cmake_version?
 
 dependency 'cmake' unless valid_cmake_version?
 dependency 'llvm'
-dependency 'tgz_package' if mac_os_x? || centos?
+dependency 'tgz_package' if macos? || mac_os_x? || centos?
```
bcardiff, line 21: ❓
If `mac_os_x?` is no longer valid for the new ominibus version we should remove them.
If they still work, what is the reason to add `macos?`

omnibus/config/software/llvm.rb
```diff
@@ -44,6 +45,7 @@
     " -DLLVM_ENABLE_ASSERTIONS=ON" \
     " -DLLVM_INCLUDE_TESTS=OFF" \
     " -DLLVM_ENABLE_Z3_SOLVER=OFF" \
+    " -DLLVM_ENABLE_LIBXML2=OFF" \
```
bcardiff, line 48: Another alternative would be to include the patches required to allow libxml2, but I am not sure we even need the xml2 actually, so I'm fine with turning it off initially.

omnibus/config/software/openssl.rb
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2012-2014 Chef Software, Inc.
+# Copyright:: Chef Software Inc.
```
bcardiff, line 2: I am skipping this. I actually think we don't need openssl. The compiler does not use it. And I think libevent_openssl is not actually used, although it is included.

So, as long as it still works I'm fine with this changes. 

Also, omnibus is currently used only on darwin, since linux builds are built with other resources.
It's good to have a chance to use it on windows, but I am not sure if that will be the path.

bcardiff (commented):
I left some comments/questions.

The next step on our side would be to build an llvm-10 again using a recent XCode (12.2+, I think) as per https://github.com/crystal-lang/distribution-scripts/pull/99#issue-662524304

This will be done updating .circleci/config.yml in this repo. Once that is done we can update the llvm_bin resources in a branch and see if that is good for maintenance build on the main repo.

All this should be done before merging this PR.

omnibus/Gemfile
maxfierke: Omnibus only publishes even number point releases I believe. Unfortunately, the last tagged is from too long ago.

omnibus/config/projects/llvm.rb
maxfierke: yeah, this was still a little confusing to me. I might need to code-dive again into the Omnibus source, but I think `macos?` is for 11.0+ and `mac_os_x` is for 10.x. I remember I was getting `false` for `mac_os_x?` initially on Big Sur.

If it's a blocker, I can try this one again tonight without it again.

omnibus/config/software/openssl.rb
maxfierke: yeah was sort of curious about that too, because we use the `make` rule to compile Crystal anyway on x86_64, which sets `-Dwithout_openssl`

maxfierke:
> The next step on our side would be to build an llvm-10 again using a recent XCode (12.2+, I think) as per [#99 (comment)](https://github.com/crystal-lang/distribution-scripts/pull/99#issue-662524304)

XCode 12.2+ I think is only needed on arm64 for linking the final binary. Shouldn't matter for Intel (unless we want to implement full cross compilation from Intel to AArch64 via Omnibus). The LLVM 10 rebuild on Intel would be more for adding the AArch64 platform to the build so the Intel builds could cross-compile for AArch64.

omnibus/config/projects/llvm.rb
bcardiff: It's not a blocker. I tried to get into the source code but found neither methods 😅. If they are needed it's fine.

bcardiff:
A new package iteration for llvm 10.0.0 with LLVM_TARGETS_TO_BUILD="X86;AArch64" has been uploaded to  https://crystal-lang.s3.amazonaws.com/llvm/llvm-10.0.0-2-darwin-x86_64.tar.gz . This was built by c63f62be4c2444d388367cf1691e5ab71455e2f7 (and I manually tweak the package iteration later 🤦 )

I am triggering a [maintenance build](https://app.circleci.com/pipelines/github/crystal-lang/crystal/6044/workflows/20c79b7b-d838-4143-8ebf-e15713bdbab6) on the compiler https://github.com/crystal-lang/crystal/commit/e38bfff6f8642ec0d0953ebe523e8329d2a2594e that we can test and accept this.

If everything goes well I might add ca523ba9d382b16fb9b1a03ca843a70bd0161f48 to this PR before merging.

FTR: 
- llvm-10.0.0-1-darwin-x86_64.tar.gz is 220M LLVM_TARGETS_TO_BUILD="X86" 
- llvm-10.0.0-2-darwin-x86_64.tar.gz is 241M LLVM_TARGETS_TO_BUILD="X86;AArch64"

bcardiff:
Ok, things need some more tweaks on paths.

I was able to get a package using llvm-10.0.0-2 in https://github.com/crystal-lang/crystal/commit/f0440a477677681bbfc0e8d2ffcdfee79b036fda and checked a 2nd generation from it in https://github.com/crystal-lang/crystal/commit/5d93d84add7643d658a30e389c057e071216d023 

I smoke tested both on Catalina 10.15.7 and everything seems fine.

- First gen: https://61614-6887813-gh.circle-artifacts.com/0/dist_packages/crystal-ci-omnibus-dev-1-darwin-x86_64.tar.gz [circle-ci job](https://app.circleci.com/pipelines/github/crystal-lang/crystal/6091/workflows/cd866a8f-d91a-4127-a84c-c16c36385b1b/jobs/61614/artifacts)
- Second gen: https://61618-6887813-gh.circle-artifacts.com/0/dist_packages/crystal-ci-omnibus-dev-1-darwin-x86_64.tar.gz [circle-ci job](https://app.circleci.com/pipelines/github/crystal-lang/crystal/6092/workflows/94d46bc3-74a1-40c2-9577-61e05db7fe2a/jobs/61618/artifacts)

Next steps:

- bump distribution-scripts in main repo
- decide which resources to use to build an official m1 package. Note that CI is not an option AFAIK.
- build an llvm 10.0.0 for M1
- decide how to build to boot/cross-compile the first m1 compiler
  a. release Crystal 1.0.0-2 x64, use it to cross-compile Crystal 1.1.0 M1 (so VERSION.x64 builds VERSION+1.aarch64)
  b. release Crystal 1.1.0 x64, use it to cross-compile Crystal 1.1.0 M1 (so VERSION.x64 builds VERSION.aarch64)
  c. release Crystal 1.0.0-2 x64, use it to cross-compile Crystal 1.0.0 M1 (same as before but starting with 1.0)
- Update homebrew formula to support m1 booting from this new compiler (no cross-compilation needed)
  - For the Crystal 1.1 in brew be built from Crystal 1.0 as it is usually done we should go with option (c), yet it probably doesn't matter much.
- Update distribution-scripts so previous m1 compiler is used to build the next version

@maxfierke you will notice some overlapping changes on the makefile with your other PRs, since the pkg had the arch suffix after the omnibus update I needed to update the Makefile also, and I apply your other changes there too.

bcardiff (approved):
Once this is merged we can apply the relevant changes from https://github.com/crystal-lang/crystal/commit/f0440a477677681bbfc0e8d2ffcdfee79b036fda . ie: bump distribution-scripts sha and ruby version used.

maxfierke:
That all makes sense and sounds good to me. One potential snag with (c) is that we'd need to put some stuff into the omnibus setup for cross-compiling `libcrystal.a` if we want to compile 1.0.0 for M1, since that release pre-dates the removal of the C-based segfault handler. But such an ad-hoc build could also be done specially too, I suppose and not need to have changes committed to the omnibus setup.

Let me know if there's anything else you need from me pre-merge (or post-merge too) in terms of testing or builds! Thank you so much for all your help and time on this!

bcardiff:
Oh, the libcrystal.a dependency is definitely something might tempt us to not do 1.0.

bcardiff:
Oh, the libcrystal.a dependency is definitely something might tempt us to not do 1.0.

PR #99, opened 2021-06-05
Allow compiling on aarch64-apple-darwin 

maxfierke:
This a follow-up to #98 and adds the actual support for compiling an omnibus build of Crystal for `aarch64-apple-darwin`.

My prior belief was that LLVM 11 was needed, but it seems only a recent XCode (12.2+, I think) with the right linker is needed and LLVM 10 seems to work fine.

The LLVM changes to add the `AArch64` target will require someone on core to update the x86_64 LLVM 10 build and add the arm64 build in S3, so I might need some assistance with that (hacked around it locally with `source path: ` to test my local builds). I'm happy to supply a build for arm64 if no one on core has access to an M1 machine.

While I'm assuming official builds will require further work on getting Circle CI, etc. setup with M1 Macs, this at least clears the way for unofficial omnibus builds by community members (such as myself) which could also be used to seed the initial ARM64 builds for Homebrew as well.

PR in draft until #98 or equivalent is merged and LLVM builds are available to update `llvm_bin` definition


darwin/Makefile
```diff
@@ -38,15 +39,15 @@ help: ## Show this help
...
+# Once there are prior builds for arm64, this can use DARWIN_ARCH
+$(CURDIR)/../omnibus/crystal-darwin-x86_64/embedded/bin/crystal:
 	curl -L -o /tmp/crystal-darwin-x86_64.tar.gz $(PREVIOUS_CRYSTAL_RELEASE_DARWIN_TARGZ) \
-	&& tar xfz /tmp/crystal-darwin-x86_64.tar.gz -O "crystal-*/embedded/bin/crystal" > $(CURDIR)/../omnibus/crystal-darwin-x86_64 \
-	&& tar xfz /tmp/crystal-darwin-x86_64.tar.gz -O "crystal-*/embedded/bin/shards" > $(CURDIR)/../omnibus/shards-darwin-x86_64 \
+	&& mkdir -p $(CURDIR)/../omnibus/crystal-darwin-x86_64
+	&& tar xfz /tmp/crystal-darwin-x86_64.tar.gz -C $(CURDIR)/../omnibus/crystal-darwin-x86_64 --strip-components=1 \
 	&& rm /tmp/crystal-darwin-x86_64.tar.gz \
-	&& chmod +x $(CURDIR)/../omnibus/crystal-darwin-x86_64 \
-	&& chmod +x $(CURDIR)/../omnibus/shards-darwin-x86_64
```
bcardiff, line 49: Note: We no longer needs shards executable. So it's good to drop the last chmod +x ✅

omnibus/config/software/crystal.rb
```diff
@@ -36,8 +37,12 @@
   env["PATH"] = "#{llvm_bin.project_dir}/bin:#{project_dir}/deps:#{env["PATH"]}"
 end
 
-if mac_os_x?
+if macos? || mac_os_x?
   env["CRYSTAL_PATH"] = "/private/var/cache/omnibus/src/crystal/src"
+
+  if arm?
+    env['CRYSTAL_CONFIG_TARGET'] = 'aarch64-apple-darwin'
```
bcardiff, line 44: This should have a 

```
# Only needed until there are previous arm64 releases
```

, right?

omnibus/config/software/crystal.rb
```diff
@@ -49,15 +54,33 @@
...
+  end
 
   block do
     raise "Could not build crystal" unless File.exists?(output_bin)
 
-    if mac_os_x?
+    if macos? || mac_os_x?
       otool_libs = `otool -L #{output_bin}`
-      if otool_libs.include?("/usr/local/lib")
+      if otool_libs.include?("/usr/local/lib") || otool_libs.include?('/opt/homebrew/lib')
```
bcardiff, line 83: 👍  just in case to prevent brew leaking into the builds

omnibus/config/software/llvm.rb
```diff
@@ -32,7 +33,7 @@
   mkdir llvm_build_dir
   command "cmake" \
     " -DCMAKE_BUILD_TYPE=MinSizeRel" \
-    " -DLLVM_TARGETS_TO_BUILD=X86" \
+    " -DLLVM_TARGETS_TO_BUILD=\"X86;AArch64\"" \
```
bcardiff, line 36: I guess this increases the size of llvm. So maybe is only needed for bootstrapping the M1 compiler, right?

I would expect to have a case `arm? ? "AArch64" : "X86"` here and have builds for llvm on both archs.

bcardiff (commented):
A couple of additional comments for the changes that are not in #98

omnibus/config/software/llvm.rb
maxfierke: Yes it's definitely needed until there's a prior ARM64 build. However, I would argue it's nice to be able to cross-compile for all supported architectures from any platform via the official builds.

But correct, this could be a conditional depending on host platform after there's an M1 compiler.

maxfierke:
would probably recommend #104 over this, but if universal builds are a no-go, I may return to this and expand to allow at least for cross-compilation from x86_64 machine (e.g. CI) and not require being on Apple Silicon

maxfierke:
Closing this out in favor of #104

PR #102, opened 2021-06-30
Add recipe for bundled tarball including static libraries

straight-shoota:
Resolves https://github.com/crystal-lang/crystal/issues/9285


linux/test.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bats
+
+@test "bundled tarball" {
+  docker run -v $(pwd)/build:/build -w /build debian /bin/sh -e -c '
+    apt update && apt install -y gcc
```
beta-ziliani, line 5: why this line is needed?

linux/bundled.dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG debian_image
+FROM ${debian_image} AS debian
+
+RUN apt-get update \
+ && apt-get install -y curl build-essential git automake libtool
+
+ENV CFLAGS="-fPIC -pipe ${release:+-O2}"
+
+ARG libpcre_version
```
beta-ziliani, line 9: Out of consistency with the rest of the file, I'll put this line below the `FROM debian AS libpcre` (feel free to ignore)

beta-ziliani (approved):
Just one little question, the rest is good to 🚀

linux/bundled.dockerfile
straight-shoota: This arg is also needed for the final stage, hence it's put here because all of the following stages derive from the base one.

linux/test.sh
straight-shoota: We need a C compiler in the image in order to link the Crystal binary.

I don't suppose we should also be shipping a C compiler in this package? That would be *all* batteries included. But the original request was about providing the libraries.

PR #103, opened 2021-06-30
Workaround for shards makefile error

straight-shoota:
Workaround for https://github.com/crystal-lang/crystal/issues/10861


linux/Dockerfile
```diff
@@ -88,6 +88,7 @@ ARG musl_target
 RUN git clone https://github.com/crystal-lang/shards \
  && cd shards \
  && git checkout ${shards_version} \
+ && touch shard.lock \
```
beta-ziliani, line 91: Perhaps it would be sensitive to add # see #103 so we have a reference for this

beta-ziliani (approved):

PR #106, opened 2021-07-20
Add release checklists

straight-shoota:
These checklists have previously been maintained in an external document. Integrating them into this repository makes it easier to discuss changes and keep them up to date, without switching contexts.

The current state represents what was used for the 1.1.0 release. There are already a number of ideas for improvement that will follow later.

### DIFF
--- a/processes/crystal-release.md
+++ b/processes/crystal-release.md
@@ -1,0 +1,103 @@
+# Crystal release process checklist
+
+1. Submit a Freeze/Release PR
+   * It should contain the expected date of the release (~two weeks after the PR is issued).
+   * It should be populated with updates to CHANGELOG.md and VERSION.
+2. Verify Maintenance CI in the parent commit of the changelog succeed
+   * This mean either the commit was the one used for nightly build or,
+   * A branch with `release/` prefix or `ci` word was pushed
+3. Check the artifacts pass the ecosystem test
+   * Grab latest copy of https://github.com/crystal-lang/test-ecosystem
+   * Override branches of projects if needed at ./scripts/10-clone-repos.sh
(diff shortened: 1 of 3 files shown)


processes/crystal-release.md
```diff
@@ -0,0 +1,103 @@
...
+21. Update carc.in / play.crystal-lang.org (by jhass)
+    Note: This is done after Arch repos are updated
+
+## Observable Helper
+
+Build changelog lines
+https://observablehq.com/d/035be530d554ccdf
+
+Check commit history
+https://observablehq.com/d/4937e5db876fe1d4
```
straight-shoota, line 103: @bcardiff These links do not appear to be working for me.

bcardiff: Yeah, they are not public. The are notebooks that use GitHub API via personal access token. I need to make some tweaks so we store those encrypted in local store. So far I've been using built-in secrets functionality but that is limited to private notebooks.

beta-ziliani: I should've mention this to you @straight-shoota , sorry! Let's just leave the steps (with some explanation maybe), but without links.

bcardiff: I can publish a version that will keep the PAT encrypted in local store. As long as each of us keep a link with their own local encryption key somewhere we are fine.

And maybe I can put some nicer urls in those 😅.

Either way. Is a living doc, it can be published and iterated.

straight-shoota: I think we might just replace these with local scripts, checked into the repository. That would be more reliable and flexible. Not sure what exactly those things do, but I already have a script that builds changelog entries from milestoned PRs. Checking all commits are milestoned shouldn't be too hard to do either.

PR #104, opened 2021-07-05
Build universal macOS builds (x86_64 & arm64)

maxfierke:
Depends on #98, supersedes #99

This is an alternative approach to Omnibus builds supporting arm64. This PR instead proposes **adding** aarch64/arm64 slices to the existing x86_64 build process to enable universal builds that can run on any Mac, Intel or Apple Silicon.

**Advantages**:

- Can compile on existing macOS-x86_64 CI infrastructure. No need for Apple Silicon build machines
- Bundles can be used on any Apple machine and can cross-compile easily for the other architecture (Linked LLVM libs and all other bundled libraries are also universal)
- Can bootstrap homebrew builds on both architectures
- Minor additions to build configs: just some extra CFLAGS and some cross-compile incantations for Crystal and Shards.

**Disadvantages**:

- Requires another LLVM rebuild (sorry Brian!) and build iteration bump
- Doubles the size of the LLVM tarball used to compile the omnibus builds
- Doubles the final distribution size: 46.7 MB (1.1.0) vs existing 22.8 MB(1.1.0)

Most of the interesting stuff is in 19fe7ff

PR #110, opened 2021-07-28
Bump llvm to 10.0.0_3 (universal darwin)

bcardiff:
This PR includes the changes of #104 to build and use llvm 10.0.0_3 that includes support for universal darwin.

In https://app.circleci.com/pipelines/github/crystal-lang/crystal?branch=ci%2Fupdate-llvm-10.0.0-3 there are builds to test this update for [first gen](https://62454-6887813-gh.circle-artifacts.com/0/dist_packages/crystal-ci-update-llvm-10.0.0-3-dev-1-darwin-x86_64.tar.gz) and [second gen](https://62491-6887813-gh.circle-artifacts.com/0/dist_packages/crystal-ci-update-llvm-10.0.0-3-dev-1-darwin-x86_64.tar.gz).

Once this is merged we can go forward and narrow the changes of #104 with the cleanup & build universal binaries.

But I think is worth bumping finally the llvm_bin package.

As noticed by @maxfierke the llvm package has doubled it size. From 200MB to 400MB. The compilers binary itself is not changing much since after this PR is not yet universal.


bcardiff:
Maybe we can also strip from #104 the cleanup of openssl and other unused patches. I think we will need to revisit eventually how the universal binaries are built and I would rather have a clean PR for that 🙈 .

jhass (approved):

straight-shoota (approved):
I'm not familiar with omnibus, but as far as I can tell it looks good.

PR #112, opened 2021-08-10
Update installer to install versioned packages

straight-shoota:
We have versioned packages now on OBS (https://build.opensuse.org/project/show/devel:languages:crystal) for each minor release. This patch adapts the installer script to install these versioned packages when the `--crystal` argument is passed.

### DIFF
--- a/packages/scripts/install.sh
+++ b/packages/scripts/install.sh
@@ -7,8 +7,8 @@ USAGE
 
     $ ./install.sh [--crystal=<crystal-version>] [--channel=stable|unstable|nightly]
 
-  - crystal-version: latest, 1.0.0, 1.1 etc. (Default: latest)
-  - channel: stable, unstable, nightly. (Default: stable)
+  - crystal-version: "latest", or a minor release version like 1.0 or 1.1 (Default: latest)
+  - channel: "stable", "unstable", "nightly" (Default: stable)
 
 REQUIREMENTS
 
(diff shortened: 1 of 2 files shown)


dentarg:
>This patch adapts the installer script to install these versioned packages when the `--version` argument is passed.

You mean `--crystal`, right?

beta-ziliani (approved):
I confirm the .deb works (Ubuntu 18.04)

straight-shoota:
@dentarg Yes, I fixed that. Maybe it should really be `--version`? I now I've made that mistake not for the first time...

beta-ziliani:
>  Maybe it should really be --version? I now I've made that mistake not for the first time...

Definitively, but it's a change that's not worth the breakage of people's own installation scripts

straight-shoota:
This argument has been broken since bintray was replaced by OBS. It only worked for `latest` or a version number matching the currently latest release. So it's unlikely to be in widespread use. (And it was only introduced about a year ago)

Replacing it doesn't need to be a breaking change. We can just add `--version` as an alternative and show a deprecation message if `--crystal` is used. This can be separate change, but if we want to change that, we should do it before announcing support for the version selector.

straight-shoota:
@bcardiff Do you remember the reason to use `--crystal` instead of `--version` in the first place?

bcardiff:
My motivation was to eventually support also --shards in case we split the deb packages. That's why I preferred --crystal instead of --version.

Also inspired in the GitHub Action that support specifying crystal and shards version independently.

bcardiff (approved):

straight-shoota:
I think we can add a separate shards version with `--shards-version` if we want to. But I think `--version` feels just more natural for a Crystal installer.

PR #113, opened 2021-08-12
Installer: Rename `--crystal` argument to `--version`

straight-shoota:
This patch introduces `--version` as a CLI argument for the installer script because that's a more intuitive name.

The previous argument `--crystal` is still usable, but triggers a deprecation message.

Suggested in https://github.com/crystal-lang/distribution-scripts/pull/112#issuecomment-897112000

### DIFF
--- a/packages/scripts/install.sh
+++ b/packages/scripts/install.sh
@@ -5,7 +5,7 @@ _help() {
 
 USAGE
 
-    $ ./install.sh [--crystal=<crystal-version>] [--channel=stable|unstable|nightly]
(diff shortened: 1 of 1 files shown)


jhass (approved):

bcardiff (approved):
When this live the post introducing the install script should be updated with a notice (or reposted) together with the update of major.minor restriction only.

straight-shoota:
@bcardiff Yes, I'm planning on writing a new announcement for this, and also update existing mentions.

PR #128, opened 2021-09-20
Revert "Merge pull request #88 from crystal-lang/fix/static-executabl…

bcardiff:
…e-check"

This reverts commit 24226f21b54709687fa4d1101a99bcdd55ed4a51.

Needed to make build pass on https://github.com/crystal-lang/crystal/pull/11228

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -80,7 +80,7 @@ RUN git clone https://github.com/crystal-lang/crystal \
  \
  && make crystal stats=true static=true ${release:+release=true} \
                  CRYSTAL_CONFIG_TARGET=${gnu_target} CRYSTAL_CONFIG_LIBRARY_PATH= \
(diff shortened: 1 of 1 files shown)


straight-shoota:
This is really strange. On 3.13 the output of `ldd crystal-static` is now again a  single line to stdout: `/lib/ld-musl-x86_64.so.1 (0x7fe96dd5d000)` (i.e. the loader), instead of an error (`Not a valid dynamic program`) printed to stderr (which was observed in #87).

The output of `file` is now
```
ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

It seems wrong that it's still a dynamically linked library for musl's linker. The GNU linker reports the same binary as statically linked.

bcardiff:
I agree it's weird. The static check can be changed if there is something better. I was only aiming to unblock the bump by checking latest changes.

straight-shoota (approved):
Sure, let's merge this to keep moving.

PR #132, opened 2021-09-21
Remove deb and rpm build instructions

straight-shoota:
The instructions for building deb and rpm packages are no longer in use and can be removed.

We're building packages on OBS now (https://github.com/crystal-lang/crystal/issues/10626).


bcardiff (approved):
🤩

PR #135, opened 2021-09-22
Add `packages/Makefile`

straight-shoota:


beta-ziliani (approved):
Note: I haven't tested it yet (but I don't expect any issue)

PR #136, opened 2021-09-22
Build ubuntu 64-bit docker image from tar.gz instead of deb

straight-shoota:
This patch changes the docker image to install Crystal directly from the generic linux tarball, instead of a debian package. The main purpose is to avoid a dependency on the debian package, which is built in a separate workflow on OBS. 

There should be no significant differences for users.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -3,13 +3,13 @@
 ## Build all docker images
 ##   $ make DOCKER_TAG=nightly CRYSTAL_VERSION=0.xy.z CRYSTAL_DEB=... CRYSTAL_TARGZ=...
 ## Build ubuntu64 docker images
-##   $ make ubuntu64 DOCKER_TAG=nightly CRYSTAL_VERSION=0.xy.z CRYSTAL_DEB=...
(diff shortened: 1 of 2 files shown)


docker/Makefile
```diff
@@ -48,10 +48,10 @@ alpine: ## Build alpine images
 alpine: $(OUTPUT_DIR)/docker-$(CRYSTAL_VERSION)-alpine.tar.gz
 alpine: $(OUTPUT_DIR)/docker-$(CRYSTAL_VERSION)-alpine-build.tar.gz
 
-$(BUILD_CONTEXT)/ubuntu32: ubuntu32.Dockerfile $(BUILD_CONTEXT)/ubuntu32/crystal.deb
+$(BUILD_CONTEXT)/ubuntu32: ubuntu32.Dockerfile $(BUILD_CONTEXT)/ubuntu32/crystal.tar.gz
```
beta-ziliani, line 51: According to line 11, 32-bit images are still using .deb, or did I miss something?

straight-shoota: Oops, yeah this change is incorrect.

beta-ziliani (approved):
🚀

PR #137, opened 2021-09-22
Add default value for CRYSTAL_DOCKER_IMAGE

straight-shoota:
Follow-up to #130 adding another default value to a Makefile argument


bcardiff:
I would argue that if there's gonna be a default it should have a the version number and not latest (or maybe even both tags)

straight-shoota:
`$(CRYSTAL_VERSION)` sounds good as a default. :+1:

bcardiff (approved):

Issue #42, opened 2019-08-30
Remove stackbottom patch


j8r:
~Proposing [feature-thread-stackbottom.patch](https://github.com/crystal-lang/distribution-scripts/blob/master/linux/files/feature-thread-stackbottom.patch) to https://github.com/ivmai/bdwgc would be a win for everybody:~
- ~Crystal core developers won't have to maintain this patch~
- ~Crystal developers won't have to recompile bdwgc to have this feature, it will be directly available in the OS package~
- ~bdwgc users will be able to use it~

Edit: the patch is already merged with https://github.com/ivmai/bdwgc/commit/5668de71107022a316ee967162bc16c10754b9ce
Remains to remove the patch when a [new release of bdwgc](https://github.com/ivmai/bdwgc/releases) will be out.

jhass:
See https://github.com/ivmai/bdwgc/pull/277

j8r:
Thanks @jhass .
I see that https://github.com/ivmai/bdwgc/commit/5668de71107022a316ee967162bc16c10754b9ce is merged @bcardiff , do you think the patch can now be removed?

bcardiff:
Until there is no bdwgc release, the only change we could do is use the patch from upstream that has a slightly different signature. It’s something on my radar. It needs to be in sync with crystal master for crystal release.

Definitely something to happen before preview_mt becomes stable.

j8r:
@jhass I updated the description, can you reopen the issue?

bcardiff:
The patch was updated with upstream in #47.

We still need this issue open to remove the patch once there is a new release of bdw-gc.

PR #141, opened 2021-10-01
Update libgc to 8.2.0

straight-shoota:
Also removes the stackbottom patch because it's included in the release.

It is yet to be determined if we want to ship 8.2.0 because it's only a pre-release version (https://github.com/crystal-lang/crystal/issues/11151#issuecomment-932198755). But it's the first release including the stackbottom patch.

Resolves https://github.com/crystal-lang/crystal/issues/11151
Resolves https://github.com/crystal-lang/distribution-scripts/issues/42


omnibus/config/software/bdw-gc.rb
```diff
@@ -20,6 +20,14 @@
   source md5: "67a5093e2f9f381bd550aa891d00b54b"
 end
 
+version "8.0.6" do
```
beta-ziliani, line 23: What's the point of keeping this one?

linux/Dockerfile
```diff
@@ -17,15 +17,12 @@ RUN apt-get install -y --allow-downgrades libcurl3-gnutls=7.64.0-4+deb10u2
 # Build libgc
 ARG gc_version
 ARG libatomic_ops_version
-COPY files/feature-thread-stackbottom-upstream.patch /tmp/
 RUN git clone https://github.com/ivmai/bdwgc \
  && cd bdwgc \
  && git checkout ${gc_version} \
  && git clone https://github.com/ivmai/libatomic_ops \
```
beta-ziliani, line 23: Not related to this PR, but I understand that libatomic_ops is not needed in 8.X versions of libgc (for modern C++14 compilers, I assume they are the default now)

omnibus/config/software/bdw-gc.rb
straight-shoota: I don't know. Same as keeping the older ones... 🤷

bcardiff: Easier to try a previous version of the gc. But I think that a cleanup is reasonable at this point.

beta-ziliani (approved):
🚀

straight-shoota:
CI workflow run: https://app.circleci.com/pipelines/github/crystal-lang/crystal/6970/workflows/9487868a-dbb5-4f68-ac7f-0a2482707f7b

PR #142, opened 2021-10-01
Add pkg-config to bundled dockerfile

straight-shoota:
Building libevent complained about missing `pkg-config`. Not sure why because I'm pretty sure it worked before. But adding it fixes that, so this seems to be the way to go.


sdogruyol (approved):
Thank you @straight-shoota 🙏

bcardiff (approved):

PR #170, opened 2021-11-18
Add libffi to build images and in the compiler build workflows

straight-shoota:
The upcoming interpreter feature requires libffi (https://github.com/crystal-lang/crystal/pull/11475).

It's only needed in the compiler, so the regular packages don't need to ship the library. It's statically linked into the compiler.

But the build images need it in order to run specs for libffi in CI.

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -44,6 +44,6 @@ FROM runtime as build
 
 RUN \
   apk add --update --no-cache --force-overwrite \
-    llvm10-dev llvm10-static g++
+    llvm10-dev llvm10-static g++ libffi-dev
 
 CMD ["/bin/sh"]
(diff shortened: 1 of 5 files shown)


straight-shoota:
After this is merged, we should rebuild and update the build images to include the library, so we can run the specs properly (see https://github.com/crystal-lang/crystal/pull/11475#issuecomment-972937107).

We should do this at least for the 1.2.2 images.
1.0.0 and 1.1.1 are also used in CI, but we could as well just skip these specs there.

beta-ziliani (approved):
🚀

straight-shoota:
We need to adopt this change to the (existing) Docker images in order to unblock running FFI specs in CI on the regular distribution images.

I think the easiest way is to update the distribution-scripts reference in the release branch (https://github.com/crystal-lang/crystal/tree/release/1.2). That should trigger a maintenance build and we can promote the resulting docker images to replace the 1.2.2 images. We could also just build those images locally, but committing to the release branch makes sense to keep this documented.

PR #178, opened 2022-01-19
Add scripts for Crystal release workflow

straight-shoota:
This patch adds two scripts that automate parts of Crystal's release workflow.

* `prepare-crystal-release` creates a tracking issue for the release in this repository (#177 was created by that script)
* `make-crystal-release` tags the git commit, pushes it it to GitHub and attaches artifacts when CI is finished (https://github.com/crystal-lang/crystal/releases/tag/1.3.2 was created by that script).

### DIFF
--- a/processes/crystal-release.md
+++ b/processes/crystal-release.md
@@ -46,17 +46,19 @@ Add an issue `Crystal release X.Y.Z` in this repo with a copy of this document. 
       * Keys can be generated at https://console.aws.amazon.com/iam/home#/security_credentials (contact a Manas admin if you don't have access).
    2. Run `make -C docs publish_docs CRYSTAL_VERSION=${VERSION}` to publish docs to `api/${VERSION}`
    3. Run `make -C docs dist-redirect_latest CRYSTAL_VERSION=${VERSION}` to apply redirect from `api/latest` to `api/${VERSION}`
-2. [ ] Publish language reference (TBD)
+2. [ ] Publish language reference
+   1. Change default branch to `release/$VERSION`
 
 ### Binary releases
 
 1. [ ] Wait for the release build in circle CI
 2. [ ] Smoke test with test-ecosystem (again)
-3. [ ] Attach build artifacts to Github release
+3. [ ] Attach build artifacts from circleci and GitHub Actions (Windows) to GitHub release
    * `crystal-*-darwin-*.tar.gz`
    * `crystal-*-linux-*.tar.gz`
    * `crystal-*.pkg`
    * `crystal-*-docs.tar.gz`
+   * `crystal.zip` (GHA) -> `crystal-$VERSION-windows-x86_64-msvc-unsupported.zip`
 4. [ ] Push changes to OBS for building linux packages
    1. Checkout https://github.com/crystal-lang/distribution-scripts and go to [`./packages`](../packages)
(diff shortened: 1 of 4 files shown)


processes/scripts/functions.sh
```diff
@@ -0,0 +1,27 @@
...
+  step_number=$(expr $step_number + 1)
+  if [ $step_number -lt $START_STEP ]; then
+    echo " \033[33m(skipped)\033[0m"
+    return
+  fi
+
+  read -r REPLY
+
+  if [ "$REPLY" != "skip" ]; then
+    eval "$command"
```
j8r, line 23: Just using `$command`(no quotes) didn't work? `command="$@"` may have to be used.

straight-shoota: I suppose it would work without quotes. But I tend to use them even if unnecessary.

processes/crystal-release.md
```diff
@@ -46,17 +47,18 @@ Add an issue `Crystal release X.Y.Z` in this repo with a copy of this document.
       * Keys can be generated at https://console.aws.amazon.com/iam/home#/security_credentials (contact a Manas admin if you don't have access).
    2. Run `make -C docs publish_docs CRYSTAL_VERSION=${VERSION}` to publish docs to `api/${VERSION}`
    3. Run `make -C docs dist-redirect_latest CRYSTAL_VERSION=${VERSION}` to apply redirect from `api/latest` to `api/${VERSION}`
-2. [ ] Publish language reference (TBD)
+2. [ ] (minor) Publish language reference
```
beta-ziliani, line 50: this is still TBD, right? (there's no mention to what to do)

processes/scripts/functions.sh
j8r: I mean, `$command`, without `eval`. It worked whith a quick local test.

j8r: I mean, `$command`, without `eval`. It worked on a quick local test.

straight-shoota: I see. Yeah, that seems to work as well. I think `eval` is maybe more expressive. Not sure if there are any relevant pros/cons for either of them.

processes/crystal-release.md
straight-shoota: Yeah. The steps are pretty clear I think, but it's not yet implemented.

beta-ziliani (approved):
🚀

processes/scripts/functions.sh
j8r: The main thing I see is that generally in dynamic languages, it is recommended to avoid `eval`.

Here, IMO, will be as expressive, shorter and will follow general recommendations 🙂 .

straight-shoota: We're in a shell script. Evaluating shell expressions happens all over the place. I don't think you can deduce anything here from `eval` in Ruby, for example.
And in that sense there's not even a conceptual difference between expanding `$command` directly and passing it to `eval`. Both run the command.

j8r: Kind of agree it will likely make not much difference here.
It is not the case here, a valid reason to avoid its use: https://google.github.io/styleguide/shellguide.html#s6.6-eval

PR #197, opened 2022-09-19
Upgrade Linux build environment to LLVM 13

### LINKED ISSUE #147: Upgrade to LLVM 13
The latest LLVM version 13.0.0 has been released yesterday. It includes relevant fixes we've been waiting for.

A regression in LLVM 11 prevented us from upgrading to LLVM 11 or 12 (https://github.com/crystal-lang/crystal/issues/10359). The fix should be available with LLVM 13.
It has been very recently discovered that LLVM 13 should also have a fix for https://github.com/crystal-lang/crystal/issues/11047 (although that's less relevant here because we don't ship Windows builds yet).

A challenge for that is that LLVM 13 is not yet available in aports edge and I'm not sure when it will be. I don't think it would be backported to Alpine 3.12, which we're still stuck with for now (https://github.com/crystal-lang/distribution-scripts/pull/127, https://github.com/crystal-lang/crystal/issues/10366). We still don't know the reasons for the issues experienced on Alpine > 3.12.

We can either pull in the `llvm-13` packages from Alpine edge whenever they are available. Or we need to build LLVM 13 manually.

If we want to link the upcoming Crystal 1.2 builds with LLVM 13, we'll probably have to resort to the latter. We want to ship them next week.

This is obviously more effort. So maybe it's not worth it considering that we can stay with LLVM 10 for now to avoid https://github.com/crystal-lang/crystal/issues/10359.
The most relevant fix in LLVM 13 at this point is really for Windows and we can already use that in CI.

Requires https://github.com/crystal-lang/crystal/issues/11277

straight-shoota:
This upgrades several components of the Linux build environment:

* alpine:3:12 -> alpine:3.16
* debian:10 -> debian:11
* bdwgc 8.2.0 -> bdwgc 8.2.2
* LLVM 11 -> LLVM 13

Green CI workflow: https://app.circleci.com/pipelines/github/crystal-lang/crystal/9794/workflows/5f762e48-3920-4242-abc1-f4c4685c0a47

Resolves #147

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -11,9 +11,6 @@ ARG release
 ENV CFLAGS="-fPIC -pipe ${release:+-O2}"
 ENV CC="clang-8"
 
-# Termporary pinning libcurl oldrelease because current (7.74.0) is broken
-RUN apt-get install -y --allow-downgrades libcurl3-gnutls=7.64.0-4+deb10u2
-
 # Build libgc
 ARG gc_version
 
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -2,9 +2,8 @@ ARG alpine_image
 ARG debian_image
 FROM ${debian_image} AS debian
 
-RUN echo "deb http://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list \
- && apt-get update \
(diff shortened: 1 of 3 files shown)


beta-ziliani (approved):
LGTM 👍

PR #219, opened 2023-01-18
Update to libpcre2

straight-shoota:
This is a part of the effort to migrate the regex engine to PCRE2 (https://github.com/crystal-lang/crystal/issues/12790).

With https://github.com/crystal-lang/crystal/pull/12978, PCRE2 becomes the default engine at which point we should ship libpcre2 in the packages where we distribute the regex library.
That's the bundled tarball and the docker images.

PR #224, opened 2023-02-16
Build compiler with PCRE2

straight-shoota:
This goes on top of #219

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -1,7 +1,7 @@
 FROM debian:11 AS debian
 
 RUN apt-get update \
- && apt-get install -y build-essential libevent-dev libpcre3-dev automake libtool pkg-config git curl llvm-13 clang-13 \
+ && apt-get install -y build-essential libevent-dev libpcre2-dev automake libtool pkg-config git curl llvm-13 clang-13 \
  && (pkg-config || true)
 
 ARG release
@@ -25,7 +25,7 @@ RUN apk add --no-cache \
       # Statically-compiled llvm
       llvm13-dev llvm13-static \
       # Static stdlib dependencies
-      zlib-static yaml-static libxml2-dev pcre-dev libevent-static \
+      zlib-static yaml-static libxml2-dev pcre2-dev libevent-static \
       # Static compiler dependencies
       libffi-dev \
       # Build tools
--- a/omnibus/config/software/crystal.rb
+++ b/omnibus/config/software/crystal.rb
@@ -13,7 +13,7 @@ else
   source git: CRYSTAL_SRC
 end
 
-dependency "pcre"
+dependency "pcre2"
 dependency "bdw-gc"
 dependency "llvm_bin" unless FIRST_RUN
 dependency "libevent"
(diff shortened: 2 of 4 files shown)


straight-shoota:
I encountered some issues while building the compiler with PCRE2 on macos:

First problem is that PCRE2 wouldn't build with JIT support. Error message: `Must target Big Sur or newer` (https://app.circleci.com/pipelines/github/crystal-lang/crystal/11185/workflows/97c658d7-e08a-4dd7-b2b4-bf9a52508d90/jobs/74155).

I have no idea how to respond to that and I couldn't even reproduce the error when I tried to build it manually (on the CI runner via SSH access). But it fails in omnibus. I suppose we might need to configure the minimum target version for macos in some way? Anyone got any ideas?

So for now I have deactivated JIT support on macos (https://github.com/crystal-lang/distribution-scripts/pull/224/commits/d10318892dbc1c21215b14531654d9f36462ca92) and then it builds.

But then the compiler that omnibus produces (first iteration/intel version) is broken and fails building the aarch64 version.
This error happens as soon as the compiler starts (it's in the option parser) and I presume this is caused by an initialization error with libpcre2.
As you can see, the debugging info doesn't give much details:
```
Using compiled compiler at .build/crystal
Invalid memory access (signal 11) at address 0x100000000
[0x10cb3e187] *Exception::CallStack::print_backtrace:Nil +39 in /private/var/cache/omnibus/src/crystal/.build/crystal
[0x10cb3e154] ~procProc(Int32, Pointer(LibC::SiginfoT), Pointer(Void), Nil)@src/signal.cr:129 +692 in /private/var/cache/omnibus/src/crystal/.build/crystal
[0x7ff816868dfd] _sigtramp +29 in /usr/lib/system/libsystem_platform.dylib
[0x10cc38ef3] *Crystal::OnceState#once<Pointer(Bool), Pointer(Void)>:(Pointer(Bool) | Nil) +131 in /private/var/cache/omnibus/src/crystal/.build/crystal
[0x10cb49f48] *OptionParser#parse_flag_definition<String>:Tuple(String, OptionParser::FlagValue) +40 in /private/var/cache/omnibus/src/crystal/.build/crystal
[0x10cb49d84] *OptionParser#on<String, String, String, &Proc(String, Nil)>:OptionParser::Handler +292 in /private/var/cache/omnibus/src/crystal/.build/crystal
[0x10c0eb411] __crystal_main +9105 in /private/var/cache/omnibus/src/crystal/.build/crystal
[0x10c0f4c04] main +68 in /private/var/cache/omnibus/src/crystal/.build/crystal
```

Again, I've tried to recreate this the compiler manually via SSH access, but then everything works fine :shrug: 

This is all super confusing. I don't have have direct access to a Mac and I'm not familiar with the platform. I also don't know much about omnibus. So probably I'm missing some detail which causes things to break, when they're expected to succeed (which I was able to verify in a manual approach).

straight-shoota:
As it happened, a very similar error appeared when using PCRE2 with the interpreter: https://github.com/crystal-lang/crystal/actions/runs/4211579122/jobs/7310019187 That was very convenient because it's much easier to debug for me. Most importantly it hinted that there's likely a general error in the lib bindings which just doesn't have any effect on some platforms (including those that I initially tested the bindings on).

The culprit was a call to `pcre2_general_context_create`. This function receives two function pointers (and a context, but that one isn't relevant). One of the function arguments was wrongly typed (`LibC::Int` instead of `LibC::SizeT`).

Fix is available in crystal-lang/crystal#13090

straight-shoota:
CI is green on https://app.circleci.com/pipelines/github/crystal-lang/crystal/11276/workflows/df04c5a2-4037-4b69-bbee-8cb1ffd52ded
So this should be good to merge once it's approved.

beta-ziliani (approved):

PR #234, opened 2023-03-21
llvm 15.0 macos build

beta-ziliani:


omnibus/config/software/llvm.rb
```diff
@@ -27,7 +37,7 @@
 
 env = with_standard_compiler_flags(with_embedded_path)
 
-llvm_build_dir = "#{build_dir}/build-llvm"
+llvm_build_dir = "#{project_dir}/#{relative_path}/build-llvm"
```
bcardiff, line 40: Isn't this change preventing older llvm packages to be built? It's ok if so. Just that we can clean up the llvm sources to have clearer expectations.

straight-shoota:
The CI workflow ran in https://app.circleci.com/pipelines/github/crystal-lang/distribution-scripts/410/workflows/5e6a98f7-d015-43e2-8a55-7ba0361a0ee3/jobs/40 and the produced artifact was uploaded to S3.
#235 uses it in `llvm_bin` and the crystal CI workflow based on that succeeds building Crystal with LLVM 15: https://app.circleci.com/pipelines/github/crystal-lang/crystal/11526/workflows/f525e1d3-a062-4602-ad56-bcd359b23ec7/jobs/75208 (at least it should, I haven't double checked myself)

straight-shoota:
Alternatively, we could consider fetching the cmake download alongside the llvm source as described here: https://github.com/crystal-lang/crystal/pull/13208#issuecomment-1479502400

PR #246, opened 2023-05-17
Update instructions for homebrew update

straight-shoota:


bcardiff:
We should have a note to manually check if the boot version needs to be updated also, right?

straight-shoota:
That's only relevant in case of a compatibility break which should be very rare. I don't think it's worth noting it in the release checklist.

beta-ziliani (approved):

beta-ziliani:
but it makes sense to add a "If the new version is incompatible with the current boot version, bump boot version"

beta-ziliani (approved):

PR #255, opened 2023-07-18
Fix: Install `asciidoctor` in linux build image

straight-shoota:
The build job complains about missing `asciidoctor` for compiling shards' man pages.
https://app.circleci.com/pipelines/github/crystal-lang/crystal/12631/workflows/68d1c081-336a-4ecc-b6e8-e4fb3852e8ab/jobs/77518

I don't know why this only became a problem now and hadn't been an issue before.
Maybe some other dependency pulled in asciidoctor previously and now doesn't anymore? But I can't tell which one that might be.

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -29,7 +29,7 @@ RUN apk add --no-cache \
       # Static compiler dependencies
       libffi-dev \
       # Build tools
-      git gcc g++ make automake libtool autoconf bash coreutils curl
+      git gcc g++ make automake libtool autoconf bash coreutils curl asciidoctor
 
 ARG release
 ENV CFLAGS="-fPIC -pipe ${release:+-O2}"


bcardiff (approved):

straight-shoota:
Ah the reason why this didn't error before is that the man pages are checked into the repo and make only rebuilds them when they are older then the corresponding `.adoc` file. That's the case currently which then results in invoking asciidoctor.

straight-shoota:
Actually, git does not preserve modification time. So it cannot be the reason.
I suppose it might perhaps be an edge case that the git checkout is not atomic and one file has a slightly different timestamp than the other? IDK.

Technically, we should not need to rebuilt the manpages. I think we can prevent this by telling `make` that the manpages are to be considered up to date.
However, I think it could also make sense to *always* rebuilt the manpages in order to get them dated to the build time, not whenever they were last edited. I think we can enforce this by telling `make` that the manpages are to be rebuilt explicitly.

These paths are pretty much opposite of each other. But either would be viable and better than the current solution in this PR which installs asciidocotor but rebuilding is not enforced.

straight-shoota:
I think "upon release" is the wrong time to check that 😅 At this point it should either be generated fresh, or taken as is (with some ensurance that it must be up to date).

I'm proposing https://github.com/crystal-lang/shards/pull/594 to check the manpages when updating the docs.

straight-shoota:
I don't think this is necessary anymore after https://github.com/crystal-lang/shards/pull/594

PR #261, opened 2023-09-27
Remove building custom `libgc` on alpine, use `gc-dev` package

straight-shoota:
I suppose at this point we do not need to build `libgc` ourselves anymore.
We should be able to use the gc package from alpine. It's up to date including the patches that we require for Crystal.

A practical effect of this change is that by default `libgc` will be linked as a dynamic library instead of a static one. This is a subtle change but I don't expect much impact for all practical purposes.
Static linking still works with `--static` of course.


ysbaddaden:
Nice!

The only breakage is in multistage dockerfiles, if you compile in stage 1 dynamically and copy the binary in stage 2, you'll have to install the `libgc` apk to fix your image. That sounds acceptable.

straight-shoota:
Yeah, if you want to copy the binary dependency free to another image, you should built it with `--static`.

beta-ziliani (approved):

PR #271, opened 2023-11-13
Update package tests to current distro versions

straight-shoota:


packages/test-install-on-docker.sh
```diff
@@ -4,4 +4,5 @@ set -eu
 
 docker_image="${1}"
 
+docker pull "$docker_image"
 docker run -e OBS_PROJECT="${OBS_PROJECT:-}" -e CRYSTAL_VERSION="${CRYSTAL_VERSION:-}" --rm -it -v $(pwd)/scripts:/scripts -v $(pwd)/support:/support $docker_image /bin/sh -c "/support/test-install.sh ${@:2}"
```
miry, line 8: I think it is the same as `docker run --pull=always`

straight-shoota: Yeah. My motivation for a separate command was that it's easier to comment out if you ever want to skip the pull 🤷
Probably not that relevant, though. `--pull=always`/`docker pull` takes ~1s longer than without.

PR #279, opened 2023-12-22
Update shards release process

straight-shoota:


bcardiff (approved):
It would be good to have examples and/or validations to ensure VERSION value starts or not with v. I assume it shouldn't start with v.

straight-shoota:
Was easy enough 😆

PR #267, opened 2024-01-04
llvm_bin is universal?

miry:
Currently the llvm_bin points to archive:
http://crystal-lang.s3.amazonaws.com/llvm/llvm-15.0.7-3-darwin-x86_64.tar.gz

As I see from changes ed5f1f97e0a67157d886dd6675902e35199a1ab3 it should  be `universal`.

Would it make sense to rename the filename and update the https://github.com/crystal-lang/distribution-scripts/blob/f2c1e13a9df73530efc4442b15e2fba90fae56dc/omnibus/config/software/llvm_bin.rb#L21 to use static filename?

Example: 

```ruby
platform = ohai['os']
# arch = ohai['kernel']['machine']
arch = "x86_64"

source url: "http://crystal-lang.s3.amazonaws.com/llvm/llvm-#{version}-#{platform}-#{arch}.tar.gz",
       md5: source_md5

```


miry:
Running CI: https://app.circleci.com/pipelines/github/miry/crystal/261/workflows/ca7c0c29-95e9-410b-a6fc-a16d282af1a0

miry:
I found there is posobility to run Action without creating a custom branch. https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

What do you think to use manual triggering, instead of keep branch build_llvm?

miry:
@straight-shoota It is ready for review.

.github/workflows/llvm.yml
```diff
@@ -0,0 +1,78 @@
+name: Darwin LLVM build
+
+on:
+  workflow_dispatch:
```
miry, line 4: I have not tested really this approach, I think it is required to be merged in master, before the button appears in Actions tab.

straight-shoota: Yeah I think it'll only show up when merged into the main branch.

PR #284, opened 2024-01-09
Fix package update script to sleep while waiting for remote changes

straight-shoota:


ysbaddaden (approved):
What are we waiting for? Any other way to know something's ready? it might still take over 10s to complete.

straight-shoota:
I'm not entirely sure, what exactly is going on. The problem here is that between `osc copypac` and `osc branchco` the repo would be cloned and a branch created, but the actual contents would be missing.
This 10s wait gives a bit more time between these two actions and allows things to finish properly.

I'm not sure if there's a better way to detect availability explicitly. We could perheps request each expected file individually via HTTP and see if it exists?. But I'm not even sure if the HTTP endpoint would be guaranteed in sync with the CLI commands.

This blunt delay usually seems to work, so I'm fine with it. For the time being these scripts are executed manually, so if anything goes wrong, it's easy to recover and try again.

PR #270, opened 2023-11-12
MacOS target 11.0

miry:
pcre2 warning message shows, that it could be compiled against 11.0

```
./src/sljit/sljitExecAllocator.c:154:2: warning: 'pthread_jit_write_protect_np' is only available on macOS 11.0 or newer [-Wunguarded-availability-new]
        pthread_jit_write_protect_np(enable_exec);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk/usr/include/pthread.h:588:6: note: 'pthread_jit_write_protect_np' has been marked as being introduced in macOS 11.0 here, but the deployment target is macOS 10.11.0
```

Bump Macos target to 11.0.

PS: It shows on my machine, and does not mean it is the same for other places.

PR #296, opened 2024-04-06
Fix macos triple

hovsater:
I've been investigating [this issue](https://github.com/crystal-lang/crystal/issues/13846) and it's related to the default target triple produced by Crystal during compilation. It seems like in order to fix a [bug](https://github.com/crystal-lang/crystal/issues/1402) back in 2015, we stripped the minimum deployment target from the target triple in this [commit](https://github.com/crystal-lang/crystal/commit/5bbf4d193aba0663b70ee3fbf750ac86b9e966b0). With the release of Xcode 15, a [new linker was introduced](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Linking) and with it came these warning messages. It appears that the new linker is stricter in which target triples it considers valid. Specifically, it looks like the minimum deployment target is required. E.g., `aarch64-apple-darwin23.3.0` is valid, while `aarch64-apple-darwin` is not.

We still need to handle the issue of normalization that's [happening here](https://github.com/crystal-lang/crystal/blob/1abc04ed04f4dcf0a2ed8b1e825bf7a0821ebab2/src/llvm.cr#L91-L109), probably by not normalizing the macOS target triple at all and I'll do that in a follow up pull request.

### DIFF
--- a/omnibus/config/software/crystal.rb
+++ b/omnibus/config/software/crystal.rb
@@ -68,14 +68,12 @@ build do
   mkdir ".build"
   copy "#{output_bin}_x86_64", ".build/crystal"
 
-  # Compile for ARM64. Apple's clang only understands arm64, LLVM uses aarch64,
-  # so we need to sub out aarch64 in our calls to Apple tools
-  env["CXXFLAGS"] << " -target arm64-apple-darwin"
   make "deps", env: env
 
-  make "crystal stats=true release=true target=aarch64-apple-darwin FLAGS=\"#{crflags}\" CRYSTAL_CONFIG_TARGET=aarch64-apple-darwin CRYSTAL_CONFIG_LIBRARY_PATH= O=#{output_path}", env: env
+  crtarget = "arm64-apple-macosx#{ENV["MACOSX_DEPLOYMENT_TARGET"]}"
+  make "crystal stats=true release=true target=#{crtarget} FLAGS=\"#{crflags}\" CRYSTAL_CONFIG_TARGET=#{crtarget} CRYSTAL_CONFIG_LIBRARY_PATH= O=#{output_path}", env: env
 
-  command "clang #{output_path}/crystal.o -o #{output_bin}_arm64 -target arm64-apple-darwin src/llvm/ext/llvm_ext.o `llvm-config --libs --system-libs --ldflags 2>/dev/null` -lstdc++ -lpcre2-8 -lgc -lpthread -levent -liconv -ldl -v", env: env
+  command "clang #{output_path}/crystal.o -o #{output_bin}_arm64 -target #{crtarget} src/llvm/ext/llvm_ext.o `llvm-config --libs --system-libs --ldflags 2>/dev/null` -lstdc++ -lpcre2-8 -lgc -lpthread -levent -liconv -ldl -v", env: env
   delete "#{output_path}/crystal.o"
 
   # Lipo them up
--- a/omnibus/config/software/shards.rb
+++ b/omnibus/config/software/shards.rb
@@ -96,9 +96,10 @@ build do
   make "clean", env: env
 
   # Build for ARM64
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       xcode: 13.4.1
     environment:
       LLVM_VERSION: 15.0.7
-      MACOSX_DEPLOYMENT_TARGET: 10.11
+      MACOSX_DEPLOYMENT_TARGET: 11.0
```
straight-shoota, line 10: **question:** Is upgrading the target version a requirement for the other changes here, or just an unrelated effect?
If the latter, we should probably extract that into its own PR.

hovsater: I bumped it based on https://github.com/crystal-lang/distribution-scripts/pull/270 since I assumed we were using `11.0` anyway. If those are unrelated, I can drop this commit.

straight-shoota: Yeah I think it's good. But I'd rather have it in a separate PR.

hovsater: Fair enough. I've dropped the commit. 🙂

straight-shoota (approved):

hovsater:
@straight-shoota would be possible to run a build with these changes to see how the default target triple comes out?

straight-shoota:
https://app.circleci.com/pipelines/github/crystal-lang/crystal/14567/workflows/6f82a2a4-06ce-42d6-9035-6b0319ab31af/jobs/82546

hovsater:
Seems like it's working! Here's the output of the generated artefact:

```console
$ crystal -v
Crystal 1.12.0-dev [2800d9462] (2024-04-06)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             LLVM: 15.0.7
Default target: aarch64-apple-macosx11.0
```

It subsequently also silences the warning messages as the default target triple now is `arm64-apple-macosx11.0` so it bypasses the [normalization process](https://github.com/crystal-lang/crystal/blob/1abc04ed04f4dcf0a2ed8b1e825bf7a0821ebab2/src/llvm.cr#L91-L109). The only reason Crystal report it as `aarch64-apple-macosx11.0` is due to the [architecture normalization](https://github.com/crystal-lang/crystal/blob/1abc04ed04f4dcf0a2ed8b1e825bf7a0821ebab2/src/compiler/crystal/codegen/target.cr#L22-L34) done in `Crystal::Codegen::Target`.

@straight-shoota like you mentioned in https://github.com/crystal-lang/crystal/issues/13846#issuecomment-2041048568, we should probably discuss whether or not it makes sense to keep the normalization done in `LLVM.default_target_triple`.

ysbaddaden:
@hovsater as far as I recall the normalization was meant to 1. make sure we use a single flag, whatever the triple name (e.g. `armhf` -> `arm`, `amd64` -> `x86_64`) and 2. generate a triple for choosing the correct LibC binding (`src/lib_c`).

Now, forcing the normalization of the actual elements was probably not a good idea. We might want to have `#normalize` and `#normalized_architecture` instead to handle the above cases, and keep the explicit triple.

hovsater:
@ysbaddaden that makes sense. We do make some normalization in `LLVM::Codegen::Target` which gets passed whatever the default target triple is. With this change, it seems to be working fine. The architecture is normalized for any consumers of the host target, while the actual triple is left alone. 🙂

straight-shoota (approved):

RX14 (approved):

straight-shoota:
FTR: This change only affects the ARM build of the compiler. It's not expected to remove linker warnings on Intel macs.

hovsater:
Yeah, you're right. We do not set the target triple explicitly for `x86_64`, so I assume we're using an appropriate target triple already. When we follow up with the pull request addressing normalization, it should sort itself out, I believe.

Issue #285, opened 2024-01-10
Stop building our own libgc


straight-shoota:
Originally, we built and distributed our own version of `libgc` because Crystal required some patches.
Those patches are long gone (#141) and we're now simply building the upstream version of libgc.

This shouldn't be our responsibility. At this point, I think we can safely drop this dependency. Distribution packages need to make sure to install the libgc package, as they already do with other dependencies.

straight-shoota:
Some older but still commonly used distributions (such as Ubuntu 22.04, Debian 11) have quite dated versions of libgc (see https://repology.org/project/boehm-gc/versions).
On these distributions, we cannot simply use the system package because it lacks support for some features that Crystal needs.
So I guess we'll have to continue distributing a custom build of libgc.

We may consider removing it from redistributed packages (such as on OBS) when an up-to-date package is available.

straight-shoota:
These are the most relevant, still supported OS releases which ship libgc < 8.2 and their EOL dates.

* Debian 11 (2026-08-31)
* OpenSUSE 15.6 (unknown, the entire 15.x branch seems to be frozen on 7.6.4)
* Ubuntu 22.04 (2027-04-01)

Ideally, we'd only ship our own `libgc` build on these systems and use the system package everywhere else.
However, I'm not sure how (or if) it's possible to have different dependency configurations depending on the distro repository.
It should work for RPM, but no idea about DEB.

PR #300, opened 2024-05-02
Drop building libgc on Linux

straight-shoota:
We don't need to ship a specially-built version of libgc anymore. This patch drops it from the regular package and moves the build instructions to the bundled package which also includes other necessary libraries.
For this reason we're still keeping `CRYSTAL_CONFIG_LIBRARY_PATH` so that it's still possible to add bundled libraries to that path.

Resolves part of #285


beta-ziliani (approved):
LGTM

PR #301, opened 2024-05-02
Use libgc from Alpine for building the compiler on Linux

straight-shoota:
We don't need a specially-built version of libgc anymore and can use the package from Alpine, just like with any other dependency.

Resolves part of https://github.com/crystal-lang/distribution-scripts/issues/285

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -25,7 +25,7 @@ RUN apk add --no-cache \
       # Statically-compiled llvm
       llvm15-dev llvm15-static \
       # Static stdlib dependencies
-      zlib-static yaml-static libxml2-static pcre2-dev libevent-static \
+      gc-dev zlib-static yaml-static libxml2-static pcre2-dev libevent-static \
       # Static compiler dependencies
       libffi-dev \
       # Build tools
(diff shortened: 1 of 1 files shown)


linux/Dockerfile
```diff
@@ -34,20 +34,12 @@ RUN apk add --no-cache \
...
- && make -j$(nproc) CFLAGS=-DNO_GETCONTEXT
-
 # This overrides default CRYSTAL_LIBRARY_PATH baked into the binary (starting with 1.2.0)
 # or configured via wrapper script (before 1.2.0) because we want to link against
-# the newly-built libraries, not the ones shipped with the bootstrap compiler.
-ENV CRYSTAL_LIBRARY_PATH=/bdwgc/.libs/
+# the regularly installed libraries, not the ones shipped with the bootstrap compiler.
+# This particularly affects libgc which was bundled upto Crystal 1.12
+ENV CRYSTAL_LIBRARY_PATH=""
+
```
beta-ziliani, line 42: Wouldn't it make more sense to simply remove this?

straight-shoota: We're using a bootstrap compiler which ships a bundled `libgc.a` in its `CRYSTAL_LIBRARY_PATH`. It's necessary to explicitly override the path to use the system library instead.
Otherwise the build would fail because the bundled library is built against glibc but we're linking musl.

beta-ziliani (approved):

PR #312, opened 2024-06-07
Apply changes to redirection rules

straight-shoota:
```console
$ make get-website-configuration | tail -n +2 | sed -e 's/1.12.2/${CRYSTAL_VERSION}/g' > aws-config.json
```

### DIFF
--- a/docs/aws-config.json
+++ b/docs/aws-config.json
@@ -8,6 +8,30 @@
     "RoutingRules": [
         {
             "Condition": {
+                "HttpErrorCodeReturnedEquals": "404",
+                "KeyPrefixEquals": "api/1"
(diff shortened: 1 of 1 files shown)


ysbaddaden:
We better not forget to add `/api/2` if we release a Crystal 2 compiler someday.

matiasgarciaisaia:
Anything else missing in order to ship this?

> We better not forget to add `/api/2` if we release a Crystal 2 compiler someday.

On the one hand, I agree with this. On the other hand, I'd expect us to have got rid of this patchwork of rules by the time Crystal's v2. But then again, the patchwork _works_, so we could add support for v2 already.

PR #313, opened 2024-06-07
Upload compressed API docs index

matiasgarciaisaia:
So S3 and CloudFront would send them compressed over the wire. For Crystal 1.12.2, the index goes from 10MB down to 1.1MB over the wire.

Fixes crystal-lang/crystal#11427


straight-shoota (approved):
LGTM.

Although I'm not sure about considering this a fix for https://github.com/crystal-lang/crystal/issues/11427. It's a more generic discussion than just affecting stdlib API hosting.

ysbaddaden (approved):

Issue #322, opened 2024-08-01
Generic dnf support


gabriel-ss:
I was setting up a pipeline today on an amazon linux 2023-based container and got caught by surprise when the crystal installer script didn't work. I found it odd since I used it dozens of times to install the compiler on older amazon linux 2 containers, so I did a little debugging and found what was happening.

The script cleverly tries to infer the correct source from the installed package manager when the information provided by the os-release is not enough. The problem with this is that it assumes that every unknown distro using rpm will manage packages with yum. When an unknown distro uses dnf (like the newer versions of AL), the script just fails. Replacing
```bash
    *)
      # If there's no dedicated repository for the distro, try to figure out
      # if the distro is apt or rpm based and use a default repository.
      _discover_distro_type

      case "$DISTRO_TYPE" in
      deb)
        DISTRO_REPO="Debian_Unstable"
        ;;
      rpm)
        DISTRO_REPO="RHEL_7"
        ;;
      *)
        _error "Unable to identify distribution type ($ID). You may specify a repository with the environment variable DISTRO_REPO"
        _error "Please, report to https://forum.crystal-lang.org/c/help-support/11"
        exit 1
        ;;
      esac
  esac
}

_discover_distro_type() {
  DISTRO_TYPE=""
  [[ $(command -v apt-get) ]] && DISTRO_TYPE="deb" && return
  [[ $(command -v yum) ]]     && DISTRO_TYPE="rpm" && return
}
```
with

```bash
    *)
      # If there's no dedicated repository for the distro, try to figure out
      # if the distro is apt, dnf or yum based and use a default repository.
      _discover_distro_type

      case "$DISTRO_TYPE" in
      apt)
        DISTRO_REPO="Debian_Unstable"
        ;;
      dnf)
        DISTRO_REPO="Fedora_Rawhide"
        ;;
      yum)
        DISTRO_REPO="RHEL_7"
        ;;
      *)
        _error "Unable to identify distribution type ($ID). You may specify a repository with the environment variable DISTRO_REPO"
        _error "Please, report to https://forum.crystal-lang.org/c/help-support/11"
        exit 1
        ;;
      esac
  esac
}

_discover_distro_type() {
  DISTRO_TYPE=""
  [[ $(command -v apt-get) ]] && DISTRO_TYPE="apt" && return
  [[ $(command -v dnf) ]]     && DISTRO_TYPE="dnf" && return
[...]

PR #325, opened 2024-08-06
Add installer test for Amazon Linux 2023

straight-shoota:
Amazon Linux is a popular Linux distribution for which we don't provide explicit packages, but it can use the fallback packages for Fedora. Adding this test ensures we'll notice if anything breaks there.

### DIFF
--- a/packages/test-install-on-docker.sh
+++ b/packages/test-install-on-docker.sh
@@ -4,4 +4,4 @@ set -eu
 
 docker_image="${1}"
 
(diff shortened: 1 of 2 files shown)


straight-shoota:
There's also a stripped down version `public.ecr.aws/lambda/provided:al2023` which would be interesting to test because it only has microdnf (see https://github.com/crystal-lang/distribution-scripts/issues/322#issuecomment-2271673454).
But trying to run this images causes an error `entrypoint requires the handler name to be the first argument` which I'm not quite sure what to do about.
@gabriel-ss any idea?

gabriel-ss:
Ah yes, the default entrypoint for those images is not bash, you have to override it with `--entrypoint /bin/bash`

ysbaddaden (approved):

PR #366, opened 2025-07-14
Move API docs redirects to CloudFront

matiasgarciaisaia:
We've hit the limits of what S3 bucket redirections allow, so we're serving the 404 page with a `200 OK` status code - breaking HTTP, and making [AI] crawlers go crazy coming up with paths.

We can now move that redirection logic to CloudFront, and be correct on our status codes.


matiasgarciaisaia:
Deployment note: we still need to "apply" the function to the current CloudFront distribution for it to take effect. Also, we'll manually drop the old S3 bucket rules.

docs/Makefile
```diff
@@ -50,16 +50,12 @@ dist-docs_versions: $(OUTPUT_DIR)/versions-$(CRYSTAL_VERSION).json ## Update `/v
 
 $(OUTPUT_DIR)/versions-$(CRYSTAL_VERSION).json: $(OUTPUT_DIR)/$(OUTPUT_DOCS_BASE_NAME).tar.gz
 
-.PHONY: dist-redirect_latest
-dist-redirect_latest: $(OUTPUT_DIR)/aws-config.json ## Apply redirect from `api/latest` to `api/$(CRYSTAL_VERSION)` in S3 (needs to be manually applied)
-	$(AWS_CLI) s3api put-bucket-website --bucket "$(AWS_BUCKET)" --website-configuration "file://$<"
+.PHONY: cloudfront-redirect_latest
+cloudfront-redirect_latest: $(OUTPUT_DIR)/aws-cloudfront-redirect.js ## Apply redirect from `api/latest` to `api/$(CRYSTAL_VERSION)` in CloudFront (needs to be manually applied)
```
straight-shoota, line 54: **suggestion:** Is there any real benefit from putting `cloudfront` in the target name? The previous one didn't mention s3 either.
I'd prefer to keep the existing name so we don't have to update the instructions.
```suggestion
.PHONY: dist-redirect_latest
dist-redirect_latest: $(OUTPUT_DIR)/aws-cloudfront-redirect.js ## Apply redirect from `api/latest` to `api/$(CRYSTAL_VERSION)` in CloudFront (needs to be manually applied)
```

docs/aws-cloudfront-redirect.js
```diff
@@ -0,0 +1,21 @@
+function handler(event) {
+    var requestPath = event.request.uri
+    var redirectUri = null
+
+    if(requestPath.startsWith('/api/latest/')) {
+      redirectUri = requestPath.replace('/api/latest/', '/api/${CRYSTAL_VERSION}/')
+    } else if(requestPath.startsWith('/api/') && !(requestPath.startsWith('/api/0') || requestPath.startsWith('/api/1'))) {
+      redirectUri = requestPath.replace('/api/', '/api/${CRYSTAL_VERSION}/')
+    }
```
straight-shoota, line 9: **issue:** I think we also need to pass through paths with the `/api/master/` prefix (for the HEAD version of the docs) and `/api/versions.json` (metadata).

straight-shoota: In the current s3 setup, we also have the following redirects which we might want to incorporate into cloudfront:

* `/api` -> `/api/` (302)
* `/api/` -> `/api/latest` (301)
* /api/index.html` -> `/api/latest` (301)
* `/api/latest` to `/api/latest/` (301)

They should probably all redirect directly to `/api/latest/` without intermediaries and all be 301s.

Then we could delete the files `index.html` and `latest` in the s3 bucket.

matiasgarciaisaia: _Oohhh!_ Good catch.

The S3 rules don't mention anything about `/api/master/` and `/api/versions.json`, and I was scratching my head trying to understand _why does this work different, then?_. But that's because S3 rules apply _when S3 resolves to a 404 error_. This CloudFront function was meant (so far, at least) to be applied _before_ hitting S3.

My first instinct is to keep this function as simple as possible: move it to _after_ the request went to S3, and only work with 404's (as we're doing now with S3); and keep the few redirects that @straight-shoota mentions that are files in the bucket (ie, not S3 "logic").

But if you have a strong reason to move all the logic to this function, I can do that too.

straight-shoota: Yeah, I was also expecting this function to go before S3.
Because there's no point in bothering S3 when we can easily tell that the request wouldn't match anything.

But I'm not determined on whether that's actually better or not. I suppose _most_ legitimate requests should not hit a 404 anyway. So maybe it's more efficient to have the rewrite only after S3.

matiasgarciaisaia: I've changed the function to be post-S3-request. I'm still scanning the whole S3 bucket to check for every redirect out there, so we have all this logic centralised in the function - and remove everything from S3, both config and empty redirection files.

ysbaddaden: I was gonna say that: `/api/master` are the nightly docs based on `master`, right?

Do we document this somewhere in human format, with a description of each main path?

matiasgarciaisaia: From [CloudFront - Restrictions on all edge functions](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/edge-function-restrictions-all.html#function-restrictions-status-codes):

> CloudFront doesn't invoke edge functions for viewer response events when the origin returns HTTP status code 400 or higher.

🤦 

So we can't say "send the request to S3, and we'll do our thing _if that fails_" 😞

Don't know why they put this limitation in place, but it changes the framing of the function. We'll need to apply the redirects before trying to hit the bucket.

ysbaddaden: Damn. Well, the rules are static _and_ small. It's doable to hardcode them.

matiasgarciaisaia:
OK, I think the function is now working OK.

I've set `https://next.crystal-lang.org/` as a CloudFront distribution matching the production one, and replaced its API docs origin with a clone of the production one - except it doesn't have the S3 redirection rules nor the "symlinks" files.

Feel free to test every URL you can think of, so we merge this and publish on production.

PR #372, opened 2025-10-10
[CI] Update xcode to 26.0.1 on circleci

straight-shoota:
xcode 15.3 is deprecated: https://circleci.com/changelog/deprecation-of-eol-xcode-versions/

I'm taking the chance to upgrade to the newest version available.

Related to https://github.com/crystal-lang/crystal/pull/16201

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   omnibus_osx_llvm:
     macos:
-      xcode: 15.3.0
+      xcode: 26.0.1
     environment:
       LLVM_VERSION: 15.0.7
       MACOSX_DEPLOYMENT_TARGET: 10.11


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 jobs:
   omnibus_osx_llvm:
     macos:
-      xcode: 15.3.0
+      xcode: 26.0.1
```
Blacksmoke16: Circle CI doesn't seem happy about this:

https://app.circleci.com/pipelines/github/crystal-lang/crystal/19342/workflows/b9b58ed2-dcf2-4620-b2d6-fc1aaf3f6471/jobs/94196

> Job was rejected because resource class macos.m1.medium.gen1, image xcode:26.0.1 is not a valid resource class

Based on https://circleci.com/docs/guides/execution-managed/using-macos/#supported-xcode-versions-silicon, it seems to use xcode 26, we'd have to move to an `m4pro.*` resource class as this one only supports up to `16.4.0`.

straight-shoota: This change isn't even running anywhere. The relevant PR is https://github.com/crystal-lang/crystal/pull/16201.
Let's continue the discussion there first (we'll probably need to come back to this as well).

PR #385, opened 2025-11-03
Drop libevent

straight-shoota:
`libevent` is no longer a required dependency on any major platform. Since [release 1.15](https://crystal-lang.org/2025/01/09/1.15.0-released/index.html#lifetime-eventloop) (https://github.com/crystal-lang/crystal/pull/14996) we're directly using the system selectors (epoll, kqueue) on Linux, macOS, and FreeBSD. Libevent is still supported, but it needs opt-in via a compile-time flag [`-Devloop=libevent`](https://crystal-lang.org/2024/11/05/lifetime-event-loop/#effects).

It's safe to assume that barely anyone uses that.
So we should be able to remove this library from our distribution artifacts, reducing dependencies and size.

With this change, libevent is still supported, but it's an optional dependency and you have to install it explicitly if you want to use it.

We may want to use libevent in stdlib specs, so I'm adding it to the `-build` images.

### DIFF
--- a/snapcraft/README.md
+++ b/snapcraft/README.md
@@ -44,7 +44,7 @@ The following are the suggested packages to be able to use the whole standard li
 
 ```
 $ sudo apt-get install gcc pkg-config git tzdata \
-                       libpcre2-dev libevent-dev libyaml-dev \
+                       libpcre2-dev libyaml-dev \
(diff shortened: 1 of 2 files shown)


ysbaddaden (approved):
Let's get this ~~merged~~ dropped!

Issue #389, opened 2025-11-20
Restructure Docker build process


straight-shoota:
While looking into adding support for aarch64 docker images, I noticed a couple of issues with the current docker build process.

* `dist_docker` builds the images, then stores them in a gzipped tarball in order to persist them in the workspace.
* The entire build step currently takes about 4 minutes. 90 seconds of that are spent for exporting the images to disk. It might be more efficient to just push the images to the registry and pull them again when necessary.
* The images are used by follow-up steps:
  - `dist_docs`: Loads the `ubuntu-build` image and uses it to build the API docs (the regular image should actually suffice)
  - `test_dist_linux_on_docker`: Loads the `ubuntu-build` image and uses it to run the entire test suite.
  - `publish_docker`: Loads all images and publishes them to docker hub
* `test_dist_linux_on_docker` itself seems very heavy (30 minutes) for little gain. A simple smoke test to make sure Crystal works should be good enough. Running the entire test suite is unlikely to fail. It could be considered an integration test for the compiler build itself (just packaged in docker). But even then it's quite a lot. Also, we're only doing this on a single platform.
* `publish_docker` depends on `test_dist_linux_on_docker` in releases, but not nightly releases. This inconsistency seems odd.

With support for aarch64 docker images, we'll be doubling the current number of docker images and their sizes.

Here's what I'd like to do in order to simplify the entire process:

1. Remove `test_dist_linux_on_docker`. We can replace it with a simple smoke test (build hello world: #391) directly in the dockerfile.
2. Publish docker images directly to the registry instead of persisting them to the workflow. `dist_docs` can pull the single image it needs.
3. Simplify the [Makefile](https://github.com/crystal-lang/distribution-scripts/blob/3503ec664796b3fd94a3c47f563aa12a93976db9/docker/Makefile). It seems too be way more complex than it needs to be. With 2. we should end up running only single [`docker build`](https://docs.docker.com/reference/cli/docker/buildx/build/) command with a couple of parameters.

PR #390, opened 2025-11-20
Build architecture-agnostic docker images

straight-shoota:
Previously, we copied a specific tarball into the docker image. This doesn't work well for multiarch builds because we need different tarballs depending on the target architecture.
With this change, we copy the appropriate tarball from a directory which may contain tarballs for multiple architectures.
This changes the CLI interface for `docker/Makefile`: It no longer receives a `CRYSTAL_TARGZ` path but a path to a directory with the tarball(s).
The refactoring also simplifies the Makefile a little bit (but more to come).

CI workflow: https://app.circleci.com/pipelines/github/crystal-lang/crystal/19768/workflows/7c7d6601-8106-44dc-83c8-ab7f49a574cd

Ref #389

### DIFF
--- a/docker/alpine.Dockerfile
+++ b/docker/alpine.Dockerfile
@@ -11,8 +11,8 @@ RUN \
     # dev tools
     make git
 
-ARG TARGETPLATFORM
-COPY crystal-${TARGETPLATFORM#linux/}.tar.gz /tmp/crystal.tar.gz
+ARG TARGETARCH
+COPY crystal-${TARGETARCH}.tar.gz /tmp/crystal.tar.gz
 
 RUN \
   tar -xz -C /usr --strip-component=1  -f /tmp/crystal.tar.gz \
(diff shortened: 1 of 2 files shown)


docker/alpine.Dockerfile
```diff
@@ -11,8 +11,8 @@ RUN \
     # dev tools
     make git
 
-ARG crystal_targz
-COPY ${crystal_targz} /tmp/crystal.tar.gz
+ARG TARGETPLATFORM
+COPY crystal-${TARGETPLATFORM#linux/}.tar.gz /tmp/crystal.tar.gz
```
ysbaddaden: We should use `TARGETARCH` instead (`amd64`, `arm64`) that skips both the OS part (`linux/`) and the variant part (e.g. `/v8`).

```suggestion
COPY crystal-${TARGETARCH}.tar.gz /tmp/crystal.tar.gz
```

PR #391, opened 2025-11-24
Add smoke tests to dockerfiles

straight-shoota:
Simple smoke tests in preparation of dropping the heavy `test_dist_linux_on_docker`


ysbaddaden:
This will save some cache in `~/.cache/crystal` in the image (must be cleaned up). Though, to be honest, I'm not sure there's much value for these smoke tests while building the images.

PR #395, opened 2025-12-01
Refactor `docker/Makefile` to push images directly to registry

straight-shoota:
As discussed in #389, exporting the docker images to disk and reimporting them in the next job is very expensive. It should be more efficient to push them directly from the builder job to the registry.

Successful workflow run: https://app.circleci.com/pipelines/github/crystal-lang/crystal/20041/workflows/bcae698e-c557-456d-ad93-b23cf4348786

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -40,7 +40,7 @@ push: push-ubuntu-build
 push: push-alpine
 push: push-alpine-build
 
-push-%: build-%
+push-%:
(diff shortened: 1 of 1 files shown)


straight-shoota:
I suppose we could expand the recipes to be less cryptic.
However, I do prefer individual recipes for each image with explicit prerequisites. When I need to build a specific image, I'd like to be able to do just that without having to build all image variants and provide all tarballs.

PR #397, opened 2025-12-08
Snap: core22 + arm64 build target + make push

ysbaddaden:
Upgrades to `core22` (Ubuntu 22.04) from `core`(16) that didn't support architectures in addition to be very outdated.

Declares the `arm64` target in addition to the `amd64` target. Both can be built on either target, but because of the target specific tarball, they must each be built separately.

Adds the `make push` target as a convenient helper.

The `Makefile` now takes explicit variables instead of implicit environment variables:

- `ARCH`: any architecture supported by crystal & snap (`amd64` and `arm64`).
- `GRADE`: the snap grade (`devel` or `stable`).
- `CHANNEL`: the snap channel (`edge` or `edge/<branch>`).

For example:

```console
$ make ARCH=amd64 CRYSTAL_TARBALL=/tmp/workspace/build/crystal-$(CRYSTAL_VERSION)-1-linux-x86_64.tar.gz GRADE=$SNAP_GRADE
$ make ARCH=arm64 CRYSTAL_TARBALL=/tmp/workspace/build/crystal-$(CRYSTAL_VERSION)-1-linux-aarch64.tar.gz GRADE=$SNAP_GRADE
$ make push CHANNEL=$SNAP_CHANNEL
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,6 @@ docs/build/
 
 darwin/build/
 
-snapcraft/snap/snapcraft.yaml
+snapcraft/snapcraft.yaml
 snapcraft/*.snap
 snapcraft/build/
--- a/snapcraft/Makefile
+++ b/snapcraft/Makefile
@@ -1,15 +1,22 @@
-OUTPUT_DIR = build
+CRYSTAL_TARBALL =
+ARCH =
+GRADE =
+CHANNEL =
 
 .PHONY: all
-all: snap/snapcraft.yaml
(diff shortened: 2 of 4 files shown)


ysbaddaden:
I added the `push` target to the `Makefile`.

CircleCI will `pack` and `upload` directly in the same job (so we avoid installing snapcraft twice) and removed the `build ` output folder. I don't think there's much point to push them as build artifacts anymore?

Issue #396, opened 2025-12-04
Alpine Docker images require AVX, causing "Illegal instruction" on older CPUs


maiha:
## Problem

The official `crystallang/crystal:*-alpine` images fail on CPUs without AVX support:

```bash
$ grep -q avx /proc/cpuinfo && echo "AVX supported" || echo "AVX not supported"
AVX not supported

$ docker run --rm crystallang/crystal:1.18.2-alpine shards --version
Illegal instruction (core dumped)
```

This affects any x86-64 CPU without AVX support (some AMD CPUs, older Intel CPUs, certain VM configurations).

## When This Started

| Crystal | Alpine | AVX instructions in `crystal` binary |
|---------|--------|--------------------------------------|
| 1.12.0 | 3.17.7 | 0 |
| **1.13.0** | **3.20.1** | **2224** ← regression |
| 1.18.2 | 3.20.8 | 2224 |

```bash
$ docker run --rm crystallang/crystal:1.18.2-alpine sh -c \
    "apk add --no-cache binutils >/dev/null && objdump -d /usr/bin/crystal | grep -cE 'vmov|vpxor|vpadd|vpmul'"
2224
```

The issue was introduced in Crystal 1.13.0 when the base image was upgraded from Alpine 3.17 to 3.20.

## Why This Happens

The Crystal build configuration hasn't changed, but the Alpine 3.20 environment (likely LLVM or GCC defaults) now generates AVX-optimized code when `--mcpu` is not explicitly specified.

The Crystal build in `distribution-scripts/linux/Dockerfile` does not specify `--mcpu` flags. Without this, LLVM defaults to optimizing for the build host's CPU capabilities, embedding AVX into `crystal` and `shards`.

## How Other Languages Handle This

| Language | Default Target | AVX | Addressed |
|----------|---------------|-----|-----------|
| **Go** | baseline x86-64 | No | 2012 |
| **Rust** | baseline x86-64 | No | 2015 |
| **Crystal** | *(not specified)* | Yes | *not yet* |

Both Go and Rust build their official binaries with baseline x86-64 (no AVX) for maximum compatibility.
## References

- **Go**: [MinimumRequirements](https://go.dev/wiki/MinimumRequirements) - Baseline x86-64 since Go 1.0 (2012)
- **Rust**: [Issue #20777](https://github.com/rust-lang/rust/issues/20777) - Changed default to x86-64 baseline (2015, before 1.0)

## Suggested Direction

[...]

PR #398, opened 2025-12-08
Fix: upgrade PCRE2 of bundled linux to 10.46

ysbaddaden:
Upgrades PCRE2 in the "bundled linux tarball" to 10.46 to avoid issue #396 in that specific tarball.


maiha:
Confirmed working on non-AVX VM. Thank you for the quick fix!

### FYI

```console
$ docker run -it --rm crystallang/crystal:1.18.2 sh -c 'crystal eval p 1'
Illegal instruction (core dumped)

$ docker run -it --rm crystallang/crystal:nightly sh -c 'crystal eval p 1'
1

$ docker run -it --rm crystallang/crystal:nightly crystal --version
Crystal 1.19.0-dev [f6a1c53e1] (2025-12-08)

LLVM: 20.1.8
Default target: x86_64-unknown-linux-gnu
```

PR #412, opened 2026-04-17
Fix: debian package changelog

ysbaddaden:
The debian package builder for testing (and unstable) now requires the changelod signoff date to use an offset, not a zone name. `UTC` was invalid as per RFC 5322 but `GMT` is also invalid.

### DIFF
--- a/packages/obs-new-minor.sh
+++ b/packages/obs-new-minor.sh
@@ -46,9 +46,9 @@ osc mv "${OLD_PACKAGE}-docs.dsc" "${PACKAGE}-docs.dsc"
 cat <<EOF > debian.changelog
 ${PACKAGE} (${VERSION}-1) stable; urgency=low
 
-* Create package for ${BASE_PACKAGE} ${VERSION}
+  * Create package for ${BASE_PACKAGE} ${VERSION}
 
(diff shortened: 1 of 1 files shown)


bcardiff (approved):

packages/obs-new-minor.sh
```diff
@@ -48,7 +48,7 @@ ${PACKAGE} (${VERSION}-1) stable; urgency=low
 
 * Create package for ${BASE_PACKAGE} ${VERSION}
 
--- Crystal Team <crystal@manas.tech>  $(LC_ALL=en_US date --utc +'%a, %-d %b %Y %T UTC')
+-- Crystal Team <crystal@manas.tech>  $(LC_ALL=en_US date --utc +'%a, %-d %b %Y %T +0000')
```
straight-shoota: We should also indent the change details as requested in https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog

```suggestion
  * Create package for ${BASE_PACKAGE} ${VERSION}

  -- Crystal Team <crystal@manas.tech>  $(LC_ALL=en_US date --utc +'%a, %-d %b %Y %T +0000')
```

ysbaddaden: I can't apply the suggestion because it wraps a deleted line... Anyway, pushed a commit.

